PR #8, opened 2019-11-21
Fix typo for config value `updateLookup`

oscartanner:
On the Source connector documentation, there is a typo, where the configuration value `updateLookup` is written without camel case.
If this value is used without camel case, the connector will not work.


rozza:
Thanks @oscartanner

PR #10, opened 2020-03-11
Add UUID in value strategy

patwakeem:
- Adds an IdStrategy that creates a UUID from a provided string in the _id field. Supports UUIDs with or without dashes.
- Adds tests.


rozza:
Tracking in: https://jira.mongodb.org/browse/KAFKA-101

rozza:
This will be added to 1.2.0 - many thanks @patwakeem  for the PR!

PR #22, opened 2020-07-06
Allow outputting in all JSON formats supported by the MongoDB driver 

Schmidt96u:
For some specific use cases at EURA NOVA, we needed to be able to output JSON in a relaxed format using `mongo-kafka`. 

To this end, we are proposing a new property (`JSON_OUTPUT_MODE`, `json.format`) to the connector configuration that manages the JSON output format.

It would give the user 4 options, as described in the [mongo-java-driver documentation](https://mongodb.github.io/mongo-java-driver/4.0/bson/extended-json/) :

* `relaxed` 
* `strict` [default] 
* `extended` 
* `shell` 

In order to maintain the current behavior of version `1.1.0` of `mongo-kafka` we have set the default mode to `strict`. 

To produce the JSON in `strict` mode, we are using the function [`BsonDocument.toJson()`](https://github.com/mongodb/mongo-java-driver/blob/3.12.x/bson/src/main/org/bson/BsonDocument.java#L824).
For the other modes we are using `JsonWriterSettings.builder()`. We did this because using `JSONMode.STRICT` with `JsonWriterSettings.builder()`  is deprecated.

As an added benefit, it should help as well when the mongoDB driver will be upgraded from `3.12` to `4.*`, since the default format for certain function of the driver will change.

If need be, you can check our [internal review](https://github.com/euranova/mongo-kafka/pull/1) of the code.

Please tell us if there is anything we need to change to be consistent with the rest of the project, or if you feel we should implement this feature differently.

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
@@ -46,6 +46,7 @@ import com.mongodb.client.model.changestream.FullDocument;
 import com.mongodb.kafka.connect.util.ConfigHelper;
 import com.mongodb.kafka.connect.util.ConnectConfigException;
 import com.mongodb.kafka.connect.util.Validators;
+import org.bson.json.JsonMode;
 
 public class MongoSourceConfig extends AbstractConfig {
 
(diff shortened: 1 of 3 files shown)


RWaltersMA:
Hi, thanks for the PR! We have a JIRA ticket for this capability https://jira.mongodb.org/browse/KAFKA-99 that is in the pipeline and currently scheduled to be done with the source schema work.  I'll defer to ross for review of your code and see if its in alignment with what we were thinking for the capability.

jbruggem:
@RWaltersMA Thanks for your reply ! Just to be sure: should we do something to notify Ross ? Should we tag him here or reference this PR in the jira ticket ?

RWaltersMA:
@rozza  and I review these PRs as part of the product planning process, kafka-99 is scheduled and we will for sure look at your PR as part of resolving that ticket.  It may need to get tweaked to conform to naming standards and other things but at initial glance its good work.  We really appreciate you taking the time to contribute to the project, we'll keep you informed with the progress.

toch:
Hi @RWaltersMA, it seems you have recently completed Kafka-99 with 80ced36c266df2eac4737c1c8a954050fa08261d. That's a good news :tada:.

I'm @jbruggem and @Schmidt96u colleague.

According to my understanding, it seems that this PR is no more relevant.

Is there anything we could keep and resubmit (considering we are restarting from the solution you have applied in 80ced36c266df2eac4737c1c8a954050fa08261d)?

rozza:
Hi @toch,

Yes I will close this PR.  The approach we took was slightly different.  In 1.3 users will be able to provide their own `JsonWriterSettings`.  This will allow total control on the outputted / conversion of bson types to Json types.

The default outputters are:

  * c.m.k.c.s.json.formatter.DefaultJson: The legacy strict json formatter.
  * c.m.k.c.s.json.formatter.ExtendedJson: The fully type safe extended json formatter.
  * c.m.k.c.s.json.formatter.SimplifiedJson: Simplified Json
    with ObjectId, Decimals, Dates and Binary values all represented as strings.

Users will also be able to provide their own `JsonWriterSettingsProvider` implementations as well.

Ross

toch:
Yep, I like the flexibility you provide with the `JsonWriterSettingsProvider`.

Thanks for the explanation :).

PR #23, opened 2020-07-16
KAFKA-120: Use statics for commonly used strings.

durran:


durran:
@rozza Seems I cannot merge these myself.

PR #24, opened 2020-07-21
Fix mispelling with tes*.

j2gg0s:
Fix mispelling with ``tesClientUri`` and ``tesTopicPrefix``.

A really small modification, so i ignore issue.
If jira's issue is necessary, please close this pr directly


rozza:
Thanks @j2gg0s - good catch

PR #25, opened 2020-08-04
Don't try to publish a source record without a topic name

rozza:
KAFKA-132

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
@@ -192,6 +192,12 @@ public class MongoSourceTask extends SourceTask {
             getTopicNameFromNamespace(
                 prefix, changeStreamDocument.getDocument("ns", new BsonDocument()));
 
+        if (topicName.isEmpty()) {
+          LOGGER.warn(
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -192,6 +192,12 @@ public void start(final Map<String, String> props) {
             getTopicNameFromNamespace(
                 prefix, changeStreamDocument.getDocument("ns", new BsonDocument()));
 
+        if (topicName.isEmpty()) {
```
durran, line 195: Should we also add a test for this logic?

rozza: I spun this work out of #27 - where I noticed some areas that needed fixes but didn't want to conflate that PR.

The topic name is missing when Invalidate events are published (dropped collections can cause this). The test speed ups in #26 highlighted this issue as it changes how the tests are processed and this ticket was added before processing on #26 then #27.

durran: Ah ok, great.

durran:
LGTM

PR #30, opened 2020-08-21
Added infer schema value support for the source connector

rozza:
KAFKA-125

https://spruce.mongodb.com/version/5f3fe3b2850e61223280d730/tasks

This PR also adds Kafka connect logical type support for Decimals, Dates, Time values and Timestamps


src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -0,0 +1,88 @@
...
+      case DOCUMENT:
+        SchemaBuilder builder = SchemaBuilder.struct();
+        bsonValue.asDocument().forEach((k, v) -> builder.field(k, inferSchema(v)));
+        builder.name(generateName(builder));
+        return builder.build();
+      case ARRAY:
+        List<BsonValue> values = bsonValue.asArray().getValues();
+        Schema firstItemSchema =
+            values.isEmpty() ? DEFAULT_INFER_SCHEMA_TYPE : inferSchema(values.get(0));
+        if (values.stream().anyMatch(bv -> inferSchema(bv) != firstItemSchema)) {
```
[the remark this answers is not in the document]
ortex: I guess here should be used  `.equals()` method, not `==`. I mean `!inferSchema(bv).equals(firstItemSchema)`

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -0,0 +1,88 @@
...
+          return SchemaBuilder.array(DEFAULT_INFER_SCHEMA_TYPE).build();
+        }
+        return SchemaBuilder.array(inferSchema(bsonValue.asArray().getValues().get(0))).build();
```
ortex, line 63: Here will be IndexOutOfBoundsException if the array is empty.

src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java
```diff
@@ -45,29 +53,56 @@
 public class SchemaAndValueProducerTest {
 
   private static final String FULL_DOCUMENT_JSON =
```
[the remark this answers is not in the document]
ortex: I suppose two more cases should be added: 
 * empty array
 * array with single but complex type

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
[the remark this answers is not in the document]
rozza: Good catch - thanks @ortex

src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java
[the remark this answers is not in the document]
rozza: Ack done.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: Good catch - done

durran (approved):

src/main/java/com/mongodb/kafka/connect/source/schema/BsonValueToSchemaAndValue.java
```diff
@@ -113,19 +120,24 @@ public SchemaAndValue toSchemaAndValue(final Schema schema, final BsonValue bson
...
+        value = bsonNumber.intValue();
+      } else if (bsonNumber.isInt64()) {
+        value = bsonNumber.longValue();
+      } else if (bsonNumber.isDouble()) {
+        value = bsonNumber.doubleValue();
+      } else {
+        value = bsonNumber.decimal128Value().bigDecimalValue();
+      }
+    } else if (bsonValue.isTimestamp()) {
+      value = bsonValue.asTimestamp().getTime();
```
jyemin, line 136: Note that BsonTimestamp.getTime returns seconds since the epoch, whereas BsonDateTime returns milliseconds since the epoch.  Should those be normalized?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonValueToSchemaAndValue.java
```diff
@@ -173,20 +202,31 @@ private SchemaAndValue stringToSchemaAndValue(final Schema schema, final BsonVal
...
-    byte[] value;
+    Object value = null;
     switch (bsonValue.getBsonType()) {
       case STRING:
         value = bsonValue.asString().getValue().getBytes(StandardCharsets.UTF_8);
         break;
       case BINARY:
         value = bsonValue.asBinary().getData();
         break;
+      case DECIMAL128:
```
jyemin, line 213: Why does this metho handle decimal128  but not other numeric types?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -0,0 +1,89 @@
...
+      case BINARY:
+        return Schema.BYTES_SCHEMA;
+      case SYMBOL:
+      case STRING:
+        return Schema.STRING_SCHEMA;
+      case OBJECT_ID:
+      case NULL:
```
jyemin, line 71: Why does null map to string?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -0,0 +1,89 @@
...
+        bsonValue.asDocument().forEach((k, v) -> builder.field(k, inferSchema(v)));
+        builder.name(generateName(builder));
+        return builder.build();
+      case ARRAY:
+        List<BsonValue> values = bsonValue.asArray().getValues();
+        Schema firstItemSchema =
+            values.isEmpty() ? DEFAULT_INFER_SCHEMA_TYPE : inferSchema(values.get(0));
+        if (values.isEmpty()
+            || values.stream().anyMatch(bv -> !Objects.equals(inferSchema(bv), firstItemSchema))) {
+          return SchemaBuilder.array(DEFAULT_INFER_SCHEMA_TYPE).build();
```
jyemin, line 62: So if there is an array with mixed types, like `[1, "one", true]` then the schema type will be array of string?  Will that force all the values to be converted to a string as well?  Is that desirable?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonValueToSchemaAndValue.java
```diff
@@ -113,19 +120,24 @@ public SchemaAndValue toSchemaAndValue(final Schema schema, final BsonValue bson
...
+    if (bsonValue.isNumber()) {
+      BsonNumber bsonNumber = bsonValue.asNumber();
+      if (bsonNumber.isInt32()) {
+        value = bsonNumber.intValue();
+      } else if (bsonNumber.isInt64()) {
+        value = bsonNumber.longValue();
+      } else if (bsonNumber.isDouble()) {
+        value = bsonNumber.doubleValue();
+      } else {
+        value = bsonNumber.decimal128Value().bigDecimalValue();
```
jyemin, line 133: Can you add a unit test for this case?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: As there is no option to support multiple types in an array, it is the least undesirable solution.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: Because there is no null type in the Kafka Schema

src/main/java/com/mongodb/kafka/connect/source/schema/BsonValueToSchemaAndValue.java
rozza: This is due to how kafka has support for big decimals, via its logical types and for big decimals they are stored as a bytes.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonValueToSchemaAndValue.java
rozza: Yes - good catch.

rozza: Done

src/main/java/com/mongodb/kafka/connect/source/schema/BsonValueToSchemaAndValue.java
rozza: That was dead code - so removed,

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
jyemin: Does this mean the connector will convert `null` to `"null"`?  Or to `""`?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
jyemin: Does this mean we will convert that array to `["1", "one", "true"]`?

rozza: Yes, String or Json representation of values is the lowest type

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: It will be whatever the string / json formatter version of null outputs. By default "null"

jyemin: I'm wondering whether it would be better to leave the field out of the schema entirely if the value is null.  What do you think?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
jyemin: Scary, but also probably uncommon

src/main/java/com/mongodb/kafka/connect/source/schema/BsonValueToSchemaAndValue.java
jyemin: Ack

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: I think its an extremely rare case, but it is potentially a value so should be included to at least indicate the field exists.

jyemin: Huh I thought this applied to any case like

   {
     "foo": null,
     ...
   }

which isn't rare.  Is that incorrect?

rozza: I think having null values in documents is a less common pattern, probably not extremely rare though!

I've updated the code, so when inferring the schema I've made BsonNull to have an optional string schema and now set the value null.

jyemin: Agreed that `null` is less common, but not rare either.

It still seems to me like leaving it out entirely is more accurate, since in most query cases in MongoDB (`$exist` being the most prominent exception to the rule) `foo: null` is equivalent to `foo` being absent from the document.  And we have no idea whether non-null values for `foo` are likely to be string rather than any other type.

jyemin (commented):
Still have a question about null handling.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: The purpose of the schema is to describe the values in the document. For that a `null` value is different to non-existing value.

jyemin: I see.  So pulling it out of the schema would require pulling it out of the source document as well, which we should not do.  Given than, I'm not sure what else we can do about the impedance mismatch between BSON and schema, beyond documenting the behavior.

jyemin (approved):
LGTM

rozza:
Merged via the cli.

PR #31, opened 2020-08-24
Sanitized the connection string in the partition map

rozza:
KAFKA-128

https://spruce.mongodb.com/version/5f43d0c4a4cf474c70593570/tasks

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
@@ -352,14 +352,30 @@ public class MongoSourceTask extends SourceTask {
     return createPartitionMap(sourceConfig, partitionMap == null);
   }
 
-  private Map<String, Object> createPartitionMap(final MongoSourceConfig sourceConfig, final boolean recreate) {
+  private Map<String, Object> createPartitionMap(
+      final MongoSourceConfig sourceConfig, final boolean recreate) {
     if (recreate) {
+      partitionMap =
+          singletonMap(NS_KEY, singletonMap(NS_KEY, createNamespaceString(sourceConfig, false)));
+    }
+    return partitionMap;
+  }
+
+  Map<String, Object> createLegacyPartitionMap(final MongoSourceConfig sourceConfig) {
+    return singletonMap(NS_KEY, createNamespaceString(sourceConfig, true));
+  }
+
+  String createNamespaceString(final MongoSourceConfig sourceConfig, final boolean legacy) {
+    if (legacy) {
+      return format(
+          "%s/%s.%s",
+          sourceConfig.getString(CONNECTION_URI_CONFIG),
+          sourceConfig.getString(DATABASE_CONFIG),
+          sourceConfig.getString(COLLECTION_CONFIG));
+    } else {
       ConnectionString connectionString = sourceConfig.getConnectionString();
       StringBuilder builder = new StringBuilder();
       builder.append(connectionString.isSrvProtocol() ? "mongodb+srv://" : "mongodb://");
-      if (connectionString.getCredential() != null) {
-        builder.append("<username>:<password>@");
-      }
       builder.append(String.join(",", connectionString.getHosts()));
       builder.append("/");
       builder.append(sourceConfig.getString(DATABASE_CONFIG));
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -345,6 +349,30 @@ String getTopicNameFromNamespace(final String prefix, final BsonDocument namespa
   }
 
   Map<String, Object> createPartitionMap(final MongoSourceConfig sourceConfig) {
+    return createPartitionMap(sourceConfig, partitionMap == null);
+  }
+
+  private Map<String, Object> createPartitionMap(final MongoSourceConfig sourceConfig, final boolean recreate) {
+    if (recreate) {
+      ConnectionString connectionString = sourceConfig.getConnectionString();
+      StringBuilder builder = new StringBuilder();
```
durran, line 358: What would you think of actually putting this logic in the Java driver itself in the future? It seems like something that could be used in multiple places at some point.

durran:
LGTM, just made a comment but would obviously take more time, but something to think about.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -116,6 +117,7 @@
   private final AtomicBoolean isCopying = new AtomicBoolean();
 
   private MongoSourceConfig sourceConfig;
+  private Map<String, Object> partitionMap;
```
jyemin, line 120: Is this field just a cache for efficiency?  If so, it doesn't seem necessary, as creating the value is not resource-intensive.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -345,6 +349,30 @@ String getTopicNameFromNamespace(final String prefix, final BsonDocument namespa
...
+    return createPartitionMap(sourceConfig, partitionMap == null);
+  }
+
+  private Map<String, Object> createPartitionMap(final MongoSourceConfig sourceConfig, final boolean recreate) {
+    if (recreate) {
+      ConnectionString connectionString = sourceConfig.getConnectionString();
+      StringBuilder builder = new StringBuilder();
+      builder.append(connectionString.isSrvProtocol() ? "mongodb+srv://" : "mongodb://");
+      if (connectionString.getCredential() != null) {
+        builder.append("<username>:<password>@");
```
jyemin, line 361: Should we just leave this out entirely?  If we're not distinguishing between different credentials, should we distinguish between the absence or presence of credentials?

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -345,6 +349,30 @@ String getTopicNameFromNamespace(final String prefix, final BsonDocument namespa
   }
 
   Map<String, Object> createPartitionMap(final MongoSourceConfig sourceConfig) {
+    return createPartitionMap(sourceConfig, partitionMap == null);
+  }
+
+  private Map<String, Object> createPartitionMap(final MongoSourceConfig sourceConfig, final boolean recreate) {
+    if (recreate) {
```
jyemin, line 356: I don't see a test for this new logic.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
jyemin: I could see a ConnectionString#getClean() method that did something similar, but the database/collection config seems specific to this use case.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
rozza: Its cached because it can be called thousands of times - one per source record produced / change stream event

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
rozza: Ideally the toString method really shouldn't leak the sensitive information, but users may rely on it providing the original string.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
rozza: Sure can do - will remove

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
rozza: Added a test

PR #34, opened 2020-09-07
Improve the handling of JSON array configs

rozza:
Includes the original error in the exception.
Tries to auto escape if parsing the initial string fails.

KAFKA-154

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java
+++ b/src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java
@@ -40,6 +40,11 @@ public final class ConfigHelper {
   private ConfigHelper() {}
 
   public static Optional<List<Document>> jsonArrayFromString(final String jsonArray) {
+    return jsonArrayFromString(jsonArray, null);
+  }
+
+  private static Optional<List<Document>> jsonArrayFromString(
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java
```diff
@@ -48,7 +53,13 @@ private ConfigHelper() {}
             Document.parse(format("{s: %s}", jsonArray)).getList("s", Document.class);
```
rozza, line 53: This is the cause of escaping problems - Using format on the string means that users had to set extra escaping in their configuration.

src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java
```diff
@@ -48,7 +53,13 @@ private ConfigHelper() {}
...
         return s.isEmpty() ? Optional.empty() : Optional.of(s);
       } catch (Exception e) {
-        throw new ConfigException("Not a valid JSON array");
+        if (originalError != null) {
```
rozza, line 56: If the value fails try doubling up on the escaping in the string, if it still fails throw the original error.

durran (approved):

jyemin (approved):
LGTM

PR #35, opened 2020-09-07
Improved error handling for list configuration options

rozza:
Also stopped duplicating / nesting of error messages

KAFKA-146

https://spruce.mongodb.com/version/5f5653e10305b94517ec4e10/tasks

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/util/ConnectConfigException.java
+++ b/src/main/java/com/mongodb/kafka/connect/util/ConnectConfigException.java
@@ -23,8 +23,11 @@ public class ConnectConfigException extends ConfigException {
   private final String name;
   private final Object value;
 
+  private final String originalMessage;
+
   public ConnectConfigException(final String name, final Object value, final String message) {
     super(name, value, message);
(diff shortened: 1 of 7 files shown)


src/main/java/com/mongodb/kafka/connect/util/ConnectConfigException.java
```diff
@@ -36,4 +39,8 @@ public String getName() {
   public Object getValue() {
     return value;
   }
+
+  public String getOriginalMessage() {
```
durran, line 43: Just a question, couldn't we just use `getMessage()` instead of creating a new method? Or is that getting modified in the super call?

durran (approved):

src/main/java/com/mongodb/kafka/connect/util/ConnectConfigException.java
rozza: Yeap correct, the inherited error messes with the message, so we have to save it.

jyemin (approved):
LGTM

PR #37, opened 2020-09-14
KAFKA-147: copy existing databases and collections by regex

ortex:
https://jira.mongodb.org/browse/KAFKA-147


.editorconfig
```diff
@@ -0,0 +1,247 @@
+root = true
```
ortex, line 1: Most parameters are default, except:

* indent_size
* tab_width
* ij_continuation_indent_size
* ij_java_names_count_to_use_import_on_demand
* ij_java_class_count_to_use_import_on_demand

ortex:
Hi @rozza 
Please, have a look

rozza:
Hi @ortex, I think this looks good, however, rather than adding two configurations a single namespace regex config would suffice.  I converted this PR to use a single config: https://github.com/rozza/mongo-kafka/tree/k147 - let me know what you think.

ortex:
> Hi @ortex, I think this looks good, however, rather than adding two configurations a single namespace regex config would suffice. I converted this PR to use a single config: https://github.com/rozza/mongo-kafka/tree/k147 - let me know what you think.

I like your idea with a single `copy.existing.namespace.regex` config param.

ortex:
I guess `COPY_EXISTING_NAMESPACE_REGEX_DOC` should be a little more detailed. For example:

```
Regex to define from which existing namespaces data should be copied from. Regex will be run against the namespace's name which is the database name and the collection name, separated by a period e.g. `database.collection`
```


And what do we do next? 
I should cherry-pick your commit or you will create another PR?

rozza:
Great news, I'll update the docs, then will merge before 1.3.0 code friends

rozza:
Closing this PR - but it is the basis for the new feature added in KAFKA-147

ortex:
Great, thanks. 
Do you have a release date for 1.3.0?

rozza:
Oct 1st is the official plan. There aren't many tickets left so the build may be in maven by the end of this week.

PR #39, opened 2020-09-17
Fix business key update strategies to use dot notation for filters

rozza:
KAFKA-155

https://spruce.mongodb.com/version/5f6374abd6d80a0640feb74a/tasks

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/ReplaceOneBusinessKeyStrategy.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/ReplaceOneBusinessKeyStrategy.java
@@ -19,6 +19,7 @@
 package com.mongodb.kafka.connect.sink.writemodel.strategy;
 
 import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.ID_FIELD;
+import static com.mongodb.kafka.connect.sink.writemodel.strategy.WriteModelHelper.flattenKeys;
 
 import org.apache.kafka.connect.errors.DataException;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/ReplaceOneBusinessKeyStrategy.java
```diff
@@ -56,4 +66,11 @@
               + " existing `_id` value in the business key then ensure `document.id.strategy.overwrite.existing=true`.");
     }
   }
+
+  @Override
+  public void configure(final MongoSinkTopicConfig configuration) {
```
rozza, line 71: This is the important change - if using a partial id strategy, then we can't rely that the created id matches the full document.  So if that is the case we flatten the filter and use dot notation.

durran: Will there be an issue when the drivers allow for "." in field names in the future?

durran (approved):

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/ReplaceOneBusinessKeyStrategy.java
rozza: Hopefully not, I'm not sure how the server will handle dot notation when querying field names with dots.

jyemin: It will treat it as a query for a nested document, e.g. 

    MongoDB Enterprise > db.test.insert({"a.b": 1})
    WriteResult({ "nInserted" : 1 })
    MongoDB Enterprise > db.test.find({"a.b" : 1})
    MongoDB Enterprise >

PR #41, opened 2020-09-22
Added `offset.partition.name` configuration.

rozza:
Allows for custom partition naming schemes.

KAFKA-158

https://spruce.mongodb.com/version/5f6a08ded1fe0717d385c0a1/tasks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
   - [KAFKA-155](https://jira.mongodb.org/browse/KAFKA-155) Fix business key update strategies to use dot notation for filters
   - [KAFKA-105](https://jira.mongodb.org/browse/KAFKA-105) Improve `errors.tolerance=all` support in the sink and source connectors.
   - [KAFKA-106](https://jira.mongodb.org/browse/KAFKA-106) Changed `max.num.retries` default to 1. A safer default especially as the driver now has retryable writes.
+  - [KAFKA-158](https://jira.mongodb.org/browse/KAFKA-158) Added `offset.partition.name` configuration, which allows for custom partitioning naming strategies.
+    Note: This can be used to start a new change stream, when an existing offset contains an invalid resume token.
 
 ## 1.2.0
   - [KAFKA-92](https://jira.mongodb.org/browse/KAFKA-92) Allow the Sink connector to use multiple tasks.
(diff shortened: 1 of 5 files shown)


durran (approved):

Gatsby-Lee:
@rozza 

Hi, 
I happen to see your commit about "offset.partition.name"

I saw the initial conversation about this change here.
( https://jira.mongodb.org/browse/KAFKA-158 )

Basically, the initial use case was to have a way to start connector without referencing previous resume token.
With this change, this use case can be archived by providing the new value for "offset.partition.name".

My use case is totally opposite.
I'd like to use the resume token where the connector stopped.
In my case ( Amazon managed Kafka ), I can't control the "offset.storage.topic" and the name of this topic is created randomly.

If I understand correctly, I can use "offset.partition.name" ( create first and set a resume token ), then I might be able to make connector pick the starting resume token.

I am testing my use case.

At the same time, I'd like to ask you if my use can make sense and can be supported by the "offset.partition.name"

Thank you

Gatsby-Lee:
@rozza 
never mind. I went through the code.
my use case is not supported with this config.

rozza:
Hi @Gatsby-Lee,

The main functionality is to continue using the last seen resume token on restart. This is stored in the last seen offset. So on restart of the connector it should restart the change stream with that resume token.  

Note: For busy clusters it is possible for resume tokens to no longer be available as they fallen off the operation log.

I'm hoping that meets your need, if not could you file a ticket in the [KAFKA](https://jira.mongodb.org/projects/KAFKA) project, so it can be triaged and reviewed?

Cheers,

Ross

Gatsby-Lee:
@rozza 

Thank you for your comment.

I'd like to explain the details that is blocking me in MSK connect ( Amazon managed Kafka connect )
When I bring up MSK connect, it automatically creates three private topics ( having prefix double-underscore ) used by the MSK connect.

* status 
* offset
* config

The offset topic is where the resumeToken is maintained while Connector runs.

Now, if I stopped the Connector or the Connector failed/stopped, I can't the new Connector with the offset topic used by previous Connector.
(Why?) MSK connect currently doesn't allow User to set the configuration, "offset.storage.topic"

( so I tried ) while MSK connect is being created, there is small window, I can create the offset topic with the expected name. ( I can see the pattern ) And, I can inject the resumeToken into the off topic.

However, it failed with this error. I am not sure how to resolve this issue.
I cloned the source code of "MongDB Source Connector" and read the details how it works. ( your commit helps me find out the starting point. )
The "MongoSourceTask" reads the resumeToken when it starts. That's why I injected it into the topic, but I got this error.

```
[Worker-09bd0dbf53789d7e5] [2022-01-12 09:09:13,365] WARN [mango-staging-pg-v0--all--resume-test1|task-0] [Producer clientId=connector-producer-mango-staging-pg-v0--all--resume-test1-0] Error while fetching metadata with correlation id 4 : {pg_v0_resume_test1.test._all=INVALID_REPLICATION_FACTOR} (org.apache.kafka.clients.NetworkClient:1119)
```

I have two questions to you.
**Q1. If I understand correctly, "offset.partition.name" helps User to override the value of Key in offset message. Therefore, User can force Connector reads resumeToken in a specific partition in a topic. right?**
( I am aware of that the partition number is determined by Hash function. )

**Q2. do you have any idea how I can overcome the limitation I have currently?**


Thank you.

PR #45, opened 2021-01-08
Added sink support for MongoDB Changestream events

rozza:
Added CDC style support for all CRUD style changestream messages.

KAFKA-51

https://spruce.mongodb.com/version/5ff8653532f41770c230379c/tasks


src/integrationTest/java/com/mongodb/kafka/connect/ChangeStreamRoundTripTest.java
```diff
@@ -0,0 +1,317 @@
...
+    insertMany(rangeClosed(140, 150), coll1, coll2);
+    coll1.deleteMany(Filters.mod("_id", 2, 0));
+    coll2.deleteMany(Filters.mod("_id", 3, 0));
+
+    assertDatabase(original, replicated);
+  }
+
+  @Test
+  @DisplayName("Ensure collection CRUD operations are replicated")
+  void testPipelineBasedUpdatesCanBeRoundTripped() {
```
rozza, line 153: Here updates are specifically tested to ensure that truncatedArrays support is added and works. This uses the test cases from the server js tests. It also ensures that the databases are as expected at various points.

src/integrationTest/java/com/mongodb/kafka/connect/mongodb/MongoKafkaTestCase.java
```diff
@@ -180,22 +182,64 @@ public MongoDatabase getDatabaseWithPostfix() {
...
-      if (retryCount != DEFAULT_MAX_RETRIES && destination.countDocuments() < expected.size()) {
-        sleep(2000);
+    retry(
+        () -> destination.countDocuments() == expected.size(),
+        () ->
+            assertIterableEquals(
+                expected, destination.find().sort(Sorts.ascending("_id")).into(new ArrayList<>())));
+  }
+
+  public void retry(final Supplier<Boolean> check, final Runnable assertion) {
```
rozza, line 230: As these tests are testing eventual consistency, I added a retry mechanism with backoff.

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTask.java
```diff
@@ -319,8 +319,8 @@ private void tryPostProcessors(
         config.getTopic());
     return records.stream()
         .map(sinkConverter::convert)
-        .map(sd -> config.getCdcHandler().flatMap(c -> c.handle(sd)))
-        .flatMap(o -> o.map(Stream::of).orElseGet(Stream::empty))
+        .map(sd -> config.getCdcHandler().map(cdc -> cdc.createWriteModels(sd)))
```
rozza, line 322: See the change in CdcHandler as to why this changed.

src/main/java/com/mongodb/kafka/connect/sink/cdc/CdcHandler.java
```diff
@@ -40,4 +44,8 @@ public MongoSinkTopicConfig getConfig() {
   }
 
   public abstract Optional<WriteModel<BsonDocument>> handle(SinkDocument doc);
+
+  public List<WriteModel<BsonDocument>> createWriteModels(final SinkDocument doc) {
```
rozza, line 48: This change is required because in the case of `truncatedArrays`, two operations are needed one to truncate the array and one to set a value.

By default the original handle method is called.

src/main/java/com/mongodb/kafka/connect/sink/cdc/mongodb/ChangeStreamHandler.java
```diff
@@ -0,0 +1,124 @@
...
+import com.mongodb.client.model.WriteModel;
+
+import com.mongodb.kafka.connect.sink.MongoSinkTopicConfig;
+import com.mongodb.kafka.connect.sink.cdc.CdcHandler;
+import com.mongodb.kafka.connect.sink.cdc.mongodb.operations.Delete;
+import com.mongodb.kafka.connect.sink.cdc.mongodb.operations.Replace;
+import com.mongodb.kafka.connect.sink.cdc.mongodb.operations.Update;
+import com.mongodb.kafka.connect.sink.converter.SinkDocument;
+
+public final class ChangeStreamHandler extends CdcHandler {
```
rozza, line 46: This class maps the change stream event type to a Change Stream operation that builds the corresponding WriteModel.

src/main/java/com/mongodb/kafka/connect/sink/cdc/mongodb/operations/Update.java
```diff
@@ -0,0 +1,75 @@
...
+  public List<WriteModel<BsonDocument>> perform(final SinkDocument doc) {
+    BsonDocument changeStreamDocument =
+        doc.getValueDoc()
+            .orElseThrow(
+                () ->
+                    new DataException("Error: value doc must not be missing for update operation"));
+
+    BsonDocument documentKey = getDocumentKey(changeStreamDocument);
+
+    if (hasFullDocument(changeStreamDocument)) {
```
rozza, line 59: If the fullDocument is available that is used rather than creating update operation(s)

src/main/java/com/mongodb/kafka/connect/sink/cdc/mongodb/operations/Update.java
```diff
@@ -0,0 +1,75 @@
...
+      return singletonList(
+          new ReplaceOneModel<>(documentKey, getFullDocument(changeStreamDocument)));
+    }
+
+    LOGGER.debug("No full document field available, creating update operation.");
+    List<WriteModel<BsonDocument>> operations = new ArrayList<>();
+
+    BsonDocument updateDescription = getUpdateDescription(changeStreamDocument);
+    if (hasTruncatedArrays(updateDescription)) {
+      operations.add(new UpdateOneModel<>(documentKey, getTruncatedArrays(updateDescription)));
```
rozza, line 70: If there are truncatedArrays also create an update operation to truncate the data.

rozza:
I'm investigating if updates with truncatedArrays can be accomplished in a single operation. If it turns out to be possible then the code will be simplified and will fall in line with the debezium cdc handlers.

jyemin (commented):
Changes all look good but I'm going to wait for the truncatedArrays discussion to resolve before approving.

rozza:
Moving into draft as the truncatedArray support will be added in the future nearer its release in MongoDB.

rozza:
truncatedArray support has been removed and will be added in a future release. This simplifys the pattern and now it follows how the debezium CDC handling is done.

src/main/java/com/mongodb/kafka/connect/sink/cdc/mongodb/operations/Update.java
jyemin: How come?  Using update operations is often more efficient.

rozza: Is it more efficient?

rozza: Happy to remove and just use updates, although that will lead to silent errors if users upgrade MongoDB but dont update the connector. Where as fullDocument is guaranteed to work in both scenarios.

jyemin: Can you explain further the update scenario that will lead to silent errors?  I think I'm missing something.

rozza: So in the future if a user upgrades their MongoDB but not their Kafka connector. If we only use the `updateDescription` and not `fullDocument` then there would be no handling for updates that change array sizes, they would silently be missed.

So I think the choices are either:

1) Use fullDocument if it exists (i.e the user has configured it in the change stream) - this will work for all scenarios
2) Only issue updates - this will work for 4.4 but fail on newer versions

I'll add some checking code so that a hard exception is thrown if there are unknown fields in the `updateDescription` document - that way there will be no silent errors.

jyemin: What did you decide to do?

rozza: I think we'll keep full support for fullDocument and if a change stream is created without it, then it will error when unexpected fields are present in the `updateDescription`

jyemin: Is the default to create the change stream with fullDocument enabled? My concern is that if it is then by default updates will be slower due to full document replacement.

rozza: Its off by default.

jyemin (approved):
LGTM!

durran (approved):

PR #46, opened 2021-01-12
Fix LazyBsonDocument clone

rozza:
Clone the LazyBsonDocument and not the unwrapped BsonDocument
As not all LazyBsonDocument instances are valid BsonDocuments
and need unwrapping.

KAFKA-180

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
@@ -57,11 +57,11 @@ public class LazyBsonDocument extends BsonDocument {
       final Type dataType,
       final BiFunction<Schema, Object, BsonDocument> converter) {
     if (record == null) {
-      throw new IllegalArgumentException("SinkRecord can not be null");
+      throw new IllegalArgumentException("record can not be null");
     } else if (dataType == null) {
       throw new IllegalArgumentException("dataType can not be null");
     } else if (converter == null) {
-      throw new IllegalArgumentException("BiFunction can not be null");
+      throw new IllegalArgumentException("converter can not be null");
     }
     this.record = record;
     this.dataType = dataType;
--- a/src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
@@ -35,13 +35,13 @@ import org.bson.BsonValue;
 public class LazyBsonDocument extends BsonDocument {
   private static final long serialVersionUID = 1L;
 
-  private transient SinkRecord record;
-  private transient Type dataType;
-  private transient BiFunction<Schema, Object, BsonDocument> converter;
+  private final transient SinkRecord record;
+  private final transient Type dataType;
+  private final transient BiFunction<Schema, Object, BsonDocument> converter;
 
   private BsonDocument unwrapped;
 
-  enum Type {
+  public enum Type {
     KEY,
     VALUE
   }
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
```diff
@@ -143,7 +145,7 @@ public String toString() {
 
   @Override
   public BsonDocument clone() {
-    return getUnwrapped().clone();
```
rozza, line 146: This is the source of the bug - there is no guarantee that unwrapped values are valid.  So create a new instance instead and unwrapping can happen lazily

src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
```diff
@@ -58,6 +58,8 @@ public LazyBsonDocument(
       final BiFunction<Schema, Object, BsonDocument> converter) {
     if (record == null) {
       throw new IllegalArgumentException("SinkRecord can not be null");
+    } else if (dataType == null) {
+      throw new IllegalArgumentException("dataType can not be null");
```
stIncMale, line 62: The neighbour exception messages mention types of method params, not names (for example, `SinkRecord` instead of `record`). It'd be better to use a consistent scheme. I believe that param names are more useful, then param types, but regardless of the scheme you chose, consistency is of the essence.

rozza: Yeah agreed - good spot.  I've updated it to use the param name, which follows the conventions over from the notNull method in the java driver.  This class is only used internally so users should never see these messages but for the sake of completeness they are added.

src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
```diff
@@ -35,13 +35,13 @@
...
-  private transient Type dataType;
-  private transient BiFunction<Schema, Object, BsonDocument> converter;
+  private final transient SinkRecord record;
+  private final transient Type dataType;
+  private final transient BiFunction<Schema, Object, BsonDocument> converter;
 
   private BsonDocument unwrapped;
 
-  enum Type {
+  public enum Type {
```
jyemin, line 44: Does this need to be public?

rozza: Yes for a test case.

jyemin: I'd prefer not to add to the API surface only in order to write a test.  Is there any alternative?

rozza: The API surface is essentially private as users cannot be expected to use classes directly.

stIncMale (approved):

src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
jyemin: Ah, that's a good point.

src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
jyemin: Sorry, one more question about this.  I'm not sure how we are using this method in the context of the project, but the javadoc for this overridden method says:

>  By convention, the object returned by this method should be independent of this object (which is being cloned).  To achieve this independence, it may be necessary to modify one or more fields of the object returned by {@code super.clone} before returning it.  Typically, this means copying any mutable objects that comprise the internal "deep structure" of the object being cloned and replacing the references to these objects with references to the copies.  If a class contains only primitive fields or references to immutable objects, then it is usually the case that no fields in the object returned by {@code super.clone} need to be modified.

This new implementation doesn't follow that convention.  Maybe it's ok because in the context of Kafka connector, all the fields of this class are effectively, if not actually, immutable?

rozza: That's correct the mutability comes from the unwrapping which generates the mutable BsonDocument - that field isn't copied.

The reason this is required is the LazyBsonDocument should only unwrap if its actually required - clone doesn't require it.

jyemin (approved):
LGTM

PR #47, opened 2021-01-12
Fixed sink validation issue including synthetic config property

rozza:
The config definition includes two synthetic properties: The topic override template property and the topic property.  When validating these should not be included in the results. This combined with https://issues.apache.org/jira/browse/KAFKA-10600 can
cause confusion to users.

KAFKA-166

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
@@ -20,7 +20,6 @@ package com.mongodb.kafka.connect.sink;
 
 import static com.mongodb.kafka.connect.sink.MongoSinkConfig.CONNECTION_URI_CONFIG;
 import static com.mongodb.kafka.connect.sink.MongoSinkConfig.TOPICS_CONFIG;
-import static com.mongodb.kafka.connect.source.MongoSourceConfig.ERRORS_LOG_ENABLE_CONFIG;
 import static com.mongodb.kafka.connect.util.ClassHelper.createInstance;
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -453,10 +453,7 @@ RateLimitSettings getRateLimitSettings() {
   static Map<String, ConfigValue> validateAll(final String topic, final Map<String, String> props) {
     String prefix = format("%s%s.", TOPIC_OVERRIDE_PREFIX, topic);
     List<String> topicOverrides =
-        props.keySet().stream()
-            .filter(k -> k.startsWith(prefix))
-            .map(k -> k.substring(0, prefix.length()))
-            .collect(Collectors.toList());
+        props.keySet().stream().filter(k -> k.startsWith(prefix)).collect(Collectors.toList());
```
jyemin, line 456: Can you explain how this fits together?  I see that the call to `map` is removed, but I'm not clear why.

rozza: That was a bug, as the map step removed the name of the actual override.
We now keep the full original property name and report on that and its validity.

rozza: To clarify some more, previously we just gave the name of the variable used (so the override prefix was removed) rather than the config the user provided.

jyemin (approved):
LGTM

durran (approved):
LGTM

PR #48, opened 2021-01-14
Configurable topic suffix for source connector

mmolimar:
Added a new config option to enable setting a custom suffix for the topic in which the messages will be sent.

KAFKA-184


rozza:
Thanks @mmolimar I used your commit as the base for the source work in #50

PR #53, opened 2021-02-09
Fixed topics.regex sink validation issue for synthetic config property

rozza:
KAFKA-195

This is linked to #47 but I missed the usecase where users declare a `topics.regex`

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
@@ -566,7 +566,6 @@ public class MongoSinkTopicConfig extends AbstractConfig {
 
   static Map<String, ConfigValue> validateRegexAll(final Map<String, String> props) {
     Map<String, ConfigValue> results = new HashMap<>();
-    AtomicBoolean containsError = new AtomicBoolean();
     Map<String, String> sinkTopicOriginals = createSinkTopicOriginals(props);
 
     CONFIG
@@ -574,9 +573,6 @@ public class MongoSinkTopicConfig extends AbstractConfig {
         .forEach(
             (k, v) -> {
               if (!IGNORED_CONFIGS.contains(k)) {
-                if (!v.errorMessages().isEmpty()) {
-                  containsError.set(true);
-                }
                 results.put(
                     k, new ConfigValue(k, v.value(), v.recommendedValues(), v.errorMessages()));
               }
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -513,12 +513,11 @@ public NamespaceMapper getNamespaceMapper() {
...
         .forEach(
             (k, v) -> {
-              if (!v.errorMessages().isEmpty()) {
-                containsError.set(true);
-              }
-
               String name = topicOverrides.contains(prefix + k) ? prefix + k : k;
               if (props.containsKey(name) && !IGNORED_CONFIGS.contains(name)) {
+                if (!v.errorMessages().isEmpty()) {
+                  containsError.set(true);
```
rozza, line 519: Only mark the error if it not being ignored.

stIncMale, line 519: [this comment does not propose a change and does not prevent merging the PR]
The fact that we need to abuse `AtomicBoolean` as a mutable boolean holder suggests that for-each loop may be a better fit here than the `forEach` method.

stIncMale (changes requested):

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
rozza: Ack

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -574,11 +573,13 @@ public NamespaceMapper getNamespaceMapper() {
         .validateAll(sinkTopicOriginals)
         .forEach(
             (k, v) -> {
-              if (!v.errorMessages().isEmpty()) {
-                containsError.set(true);
+              if (!IGNORED_CONFIGS.contains(k)) {
+                if (!v.errorMessages().isEmpty()) {
+                  containsError.set(true);
```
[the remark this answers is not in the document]
rozza: Copy pasta - removed.

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -525,7 +520,7 @@ public NamespaceMapper getNamespaceMapper() {
               }
             });
 
-    if (!containsError.get()) {
+    if (results.values().stream().allMatch(v -> v.errorMessages().isEmpty())) {
```
rozza, line 523: As a for loop over entrysets is ugly and harder to read than a forEach loop - I opted to do a value check at the end.  Given validation is a startup process there is little overhead for this.

stIncMale (approved):

PR #56, opened 2021-02-10
Create DeleteOneBusinessKeyStrategy.java

rozza:
Create a custom delete write strategy.

KAFKA-168

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
@@ -21,8 +21,8 @@ import static com.mongodb.kafka.connect.sink.writemodel.strategy.WriteModelHelpe
 
 import org.apache.kafka.connect.errors.DataException;
 
-import org.bson.BSONException;
 import org.bson.BsonDocument;
+import org.bson.BsonValue;
 
 import com.mongodb.client.model.DeleteOneModel;
 import com.mongodb.client.model.WriteModel;
@@ -48,19 +48,19 @@ public class DeleteOneBusinessKeyStrategy implements WriteModelStrategy, Configu
                     new DataException(
                         "Could not build the WriteModel,the value document was missing unexpectedly"));
 
-    try {
-      BsonDocument businessKey = vd.getDocument(ID_FIELD);
-      vd.remove(ID_FIELD);
-      if (isPartialId) {
-        businessKey = flattenKeys(businessKey);
-      }
-      return new DeleteOneModel<>(businessKey);
-
-    } catch (BSONException e) {
+    BsonValue idValue = vd.get(ID_FIELD);
+    if (idValue == null || !idValue.isDocument()) {
       throw new DataException(
           "Could not build the WriteModel, the value document does not contain an _id field of"
               + " type BsonDocument which holds the business key fields.");
     }
+
+    BsonDocument businessKey = idValue.asDocument();
+    vd.remove(ID_FIELD);
+    if (isPartialId) {
+      businessKey = flattenKeys(businessKey);
+    }
+    return new DeleteOneModel<>(businessKey);
   }
 
   @Override
--- a/src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.kafka.connect.sink.writemodel.strategy;
+
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.ID_FIELD;
+import static com.mongodb.kafka.connect.sink.writemodel.strategy.WriteModelHelper.flattenKeys;
+
+import org.apache.kafka.connect.errors.DataException;
+
+import org.bson.BSONException;
+import org.bson.BsonDocument;
+
+import com.mongodb.client.model.DeleteOneModel;
+import com.mongodb.client.model.WriteModel;
+
+import com.mongodb.kafka.connect.sink.Configurable;
+import com.mongodb.kafka.connect.sink.MongoSinkTopicConfig;
+import com.mongodb.kafka.connect.sink.converter.SinkDocument;
+import com.mongodb.kafka.connect.sink.processor.id.strategy.IdStrategy;
+import com.mongodb.kafka.connect.sink.processor.id.strategy.PartialKeyStrategy;
+import com.mongodb.kafka.connect.sink.processor.id.strategy.PartialValueStrategy;
+
+public class DeleteOneBusinessKeyStrategy implements WriteModelStrategy, Configurable {
+
+  private boolean isPartialId = false;
+
+  @Override
+  public WriteModel<BsonDocument> createWriteModel(final SinkDocument document) {
+    BsonDocument vd =
+        document
+            .getValueDoc()
+            .orElseThrow(
+                () ->
+                    new DataException(
+                        "Could not build the WriteModel,the value document was missing unexpectedly"));
+
+    try {
+      BsonDocument businessKey = vd.getDocument(ID_FIELD);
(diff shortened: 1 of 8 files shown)


src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/ReplaceOneDefaultStrategy.java
```diff
@@ -44,6 +45,12 @@
                     new DataException(
                         "Could not build the WriteModel,the value document was missing unexpectedly"));
 
-    return new ReplaceOneModel<>(new BsonDocument(ID_FIELD, vd.get(ID_FIELD)), vd, REPLACE_OPTIONS);
+    BsonValue idValue = vd.get(ID_FIELD);
+    if (idValue == null) {
```
rozza, line 49: When refactoring the error sceanrios I noticed that these errors would cause IllegalArgumentExceptions when creating the WriteModels - so I normalized them to DataExceptions.

src/test/java/com/mongodb/kafka/connect/sink/writemodel/strategy/WriteModelStrategyTest.java
```diff
@@ -97,17 +109,6 @@
   private static final BsonDocument BUSINESS_KEY_FLATTENED_FILTER =
       BsonDocument.parse("{'a.a1': 1, 'b.b1': 1, 'b.b2': 1}");
 
-  @Test
-  @DisplayName("when key document is missing for DeleteOneDefaultStrategy then DataException")
```
rozza, line 101: Moved all error scenarios to a single test case.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/UpdateOneTimestampsStrategy.java
```diff
@@ -48,10 +49,15 @@
...
                         "Could not build the WriteModel,the value document was missing unexpectedly"));
 
-    BsonDateTime dateTime = new BsonDateTime(Instant.now().toEpochMilli());
+    BsonValue idValue = vd.get(ID_FIELD);
+    if (idValue == null) {
+      throw new DataException(
+          "Could not build the WriteModel,the `_id` field was missing unexpectedly");
+    }
 
+    BsonDateTime dateTime = new BsonDateTime(Instant.now().toEpochMilli());
```
stIncMale, line 58: If only milliseconds precision (a.k.a., resolution - the "fineness" to which a timer can be read) is needed, using `System.currentTimeMillis` instead of `Instant.now().toEpochMilli()` is more efficient because
- a new object is not created;
- a less accurate (a.k.a., more granular; granularity - the average time in for which a timer value stays the same) clock may be used under the hood (this seems to be the case in OpenJDK JDK).

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+    try {
+      BsonDocument businessKey = vd.getDocument(ID_FIELD);
+      vd.remove(ID_FIELD);
+      if (isPartialId) {
+        businessKey = flattenKeys(businessKey);
+      }
+      return new DeleteOneModel<>(businessKey);
+
+    } catch (BSONException e) {
+      throw new DataException(
```
stIncMale, line 60: It appears that there are too many statements between `vd.getDocument(ID_FIELD)` and `catch (BSONException e)` preventing us to state that any `BSONException` caught was thrown by the `vd.getDocument(ID_FIELD)` statement.

I propose to shrink the `try` block so that it contains only `vd.getDocument(ID_FIELD)`.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+          "Could not build the WriteModel,the value document does not contain an _id field of"
+              + " type BsonDocument which holds the business key fields.`.");
```
stIncMale, line 62: ``.`.``  does not seem intentional. We also probably need a space after the comma in `WriteModel,the`.

P.S. I did not tag this comment as "Nit" because it is about an error message intended to be read by users.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+
+  @Override
+  public void configure(final MongoSinkTopicConfig configuration) {
+    IdStrategy idStrategy = configuration.getIdStrategy();
+    isPartialId =
+        idStrategy instanceof PartialKeyStrategy || idStrategy instanceof PartialValueStrategy;
```
stIncMale, line 70: I am blatantly unfamiliar with the semantics of `IdStrategy`, `WriteModelStrategy`, and do not know whether or not users are allowed to create their own implementations of `IdStrategy`. But if I were to judge based only on this code, I would say that it suggests the need for either a new method `IdStrategy.isPartial`, or a marker interface `IdStrategy.Partial`, or a similar mechanism that does not require modifying this code when a new partial `IdStrategy` implementation is born.

stIncMale (changes requested):

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
rozza: That would be possible, if ever a new partial strategy is added then that would be a good time to add it.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
rozza: Agreed - will do as a separate commit as this is a pattern is repeated in all the BusinessKey based write strategies

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
rozza: Done

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
rozza: Done - I've removed the try catch and bought it inline with the other checks idValue checks added.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/UpdateOneTimestampsStrategy.java
rozza: Nice - done.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneBusinessKeyStrategy.java
stIncMale: What concerns me in this approach is that the future you creating a new partial `IdStrategy` will have to remember that this place in the codebase needs to be updated. And if it's not you, but someone else, how will that someone know that this place must also be updated?

If we had [sealed classes](https://openjdk.java.net/jeps/360), we could have left a comment near the set of all allowed subclasses about updating this place in code. Not a good solution, but at least something. However, we do not have that language feature yet (and will not have it in the driver code for the foreseeable future).

rozza: Thats still fine, its a limitation of the current implementations.Users can create their own implementations of both WriteModelStrategy (where this initial code came from) and IdStrategies. 

It hasn't been an issue for the alternative BusinessKeyStrategies which have existed for the years of this project.

stIncMale: Is my understanding correct that the current argument goes like the following?
1. Other code in the codebase, e.g., `ReplaceOneBusinessKeyStrategy` has the same (objective) issue.
2. It has been fine so far.
3. Therefore, we may continue writing code that has the issue.

If yes, then I can neither consider this as a good argument nor agree with it because
- the change needed to avoid the issue in the code is neither disruptive nor too laborious (if I am wrong about this part, please tell me);
- the fact that the same mechanism of differentiating partial `IdStrategy` is needed in other places in the codebase emphasizes the need to have this mechanism instead of repeating the same problematic workaround.

I don't know how we approach situations when the author and a reviewer cannot come to an agreement in a timely manner. Here is what I propose (again, provided that my understanding of your current argument I described above is correct):
- create ("split from" KAFKA-168) a `KAFKA` improvement about introducing a mechanism of differentiating partial `IdStrategy`;
- approve and merge the current PR.

stIncMale: Created [KAFKA-199 Introduce a mechanism of differentiating partial `IdStrategy`](https://jira.mongodb.org/browse/KAFKA-199).

PR #59, opened 2021-03-08
Improve copy.existing

rozza:
Use alternative namespace field name to prevent clashes.

KAFKA-201

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
@@ -65,7 +65,7 @@ import com.mongodb.client.MongoClient;
  */
 class MongoCopyDataManager implements AutoCloseable {
   private static final Logger LOGGER = LoggerFactory.getLogger(MongoCopyDataManager.class);
-  private static final String NAMESPACE_FIELD = "__";
+  static final String NAMESPACE_FIELD = "__";
   private static final byte[] NAMESPACE_BYTES = "ns".getBytes(StandardCharsets.UTF_8);
 
   private static final String PIPELINE_TEMPLATE =
@@ -74,7 +74,7 @@ class MongoCopyDataManager implements AutoCloseable {
               + "{newRoot: {"
               + "_id: {_id: '$_id', copyingData: true}, "
               + "operationType: 'insert', "
-              + "'%s': {db: '%%s', coll: '%%s'}"
+              + "%s: {db: '%%s', coll: '%%s'}"
               + "documentKey: {_id: '$_id'}, "
               + "fullDocument: '$$ROOT'}}"
               + "}",
--- a/src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
@@ -21,9 +21,12 @@ import static com.mongodb.kafka.connect.source.MongoSourceConfig.COPY_EXISTING_N
 import static com.mongodb.kafka.connect.source.MongoSourceConfig.COPY_EXISTING_PIPELINE_CONFIG;
 import static com.mongodb.kafka.connect.source.MongoSourceConfig.COPY_EXISTING_QUEUE_SIZE_CONFIG;
 import static com.mongodb.kafka.connect.source.MongoSourceConfig.DATABASE_CONFIG;
+import static java.lang.String.format;
 import static java.util.Collections.singletonList;
 import static java.util.stream.Collectors.toList;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -32,7 +35,6 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
@@ -40,7 +42,9 @@ import org.apache.kafka.connect.errors.ConnectException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.bson.BsonBinaryReader;
 import org.bson.BsonDocument;
+import org.bson.BsonType;
 import org.bson.RawBsonDocument;
 import org.bson.conversions.Bson;
 
@@ -61,6 +65,21 @@ import com.mongodb.client.MongoClient;
  */
 class MongoCopyDataManager implements AutoCloseable {
   private static final Logger LOGGER = LoggerFactory.getLogger(MongoCopyDataManager.class);
+  private static final String NAMESPACE_FIELD = "__";
+  private static final byte[] NAMESPACE_BYTES = "ns".getBytes(StandardCharsets.UTF_8);
+
+  private static final String PIPELINE_TEMPLATE =
+      format(
+          "{$replaceRoot: "
+              + "{newRoot: {"
+              + "_id: {_id: '$_id', copyingData: true}, "
+              + "operationType: 'insert', "
+              + "'%s': {db: '%%s', coll: '%%s'}"
+              + "documentKey: {_id: '$_id'}, "
+              + "fullDocument: '$$ROOT'}}"
+              + "}",
+          NAMESPACE_FIELD);
+
   private volatile boolean closed;
   private volatile Exception errorException;
   private final AtomicInteger namespacesToCopy;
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
```diff
@@ -61,6 +65,21 @@
...
+  private static final String NAMESPACE_FIELD = "__";
+  private static final byte[] NAMESPACE_BYTES = "ns".getBytes(StandardCharsets.UTF_8);
+
+  private static final String PIPELINE_TEMPLATE =
+      format(
+          "{$replaceRoot: "
+              + "{newRoot: {"
+              + "_id: {_id: '$_id', copyingData: true}, "
+              + "operationType: 'insert', "
+              + "'%s': {db: '%%s', coll: '%%s'}"
```
stIncMale, line 77: I have multiple questions about using quotes in strings that we call JSON (these questions are relevant not only to this PR):
1. Why are names not always encompassed in double quotes (`"`)? Both [JSON](https://www.rfc-editor.org/rfc/rfc8259.html) and [extended JSON](https://github.com/stIncMale/specifications/blob/master/source/extended-json.rst) require this.
2. Why is the namespace field name in single quotes?
3. Why are values encompassed in single quotes instead of double quotes? This, again, seem to violate both specifications.

src/test/java/com/mongodb/kafka/connect/source/MongoCopyDataManagerTest.java
```diff
@@ -436,6 +426,22 @@ private void sleep() {
...
+  private static String createTemplate(final int id) {
+    return createTemplate(id, "myDB", "myColl");
+  }
+
+  private static String createTemplate(final int id, final String dbName, final String collName) {
+    return format(CHANGE_STREAM_DOCUMENT_TEMPLATE, id, dbName, collName, id, id, id);
+  }
+
+  private static RawBsonDocument createInput(final String json) {
+    return RawBsonDocument.parse(format(json, "__"));
```
stIncMale, line 438: Should `MongoCopyDataManager.NAMESPACE_FIELD` be used here? (requires making the field package-private)

src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
```diff
@@ -166,20 +185,32 @@ private void putToQueue(final BsonDocument bsonDocument) {
...
+    ByteBuffer sourceBuffer = original.getByteBuffer().asNIO();
+    BsonBinaryReader reader = new BsonBinaryReader(sourceBuffer);
+    int currentPosition = 0;
+    reader.readStartDocument();
+    while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
+      if (reader.readName().equals(NAMESPACE_FIELD)) {
+        currentPosition++; // Adjust the current position to include the bson type
+        byte[] sourceBytes = sourceBuffer.array();
+        // Convert the namespace field in situ
+        for (byte namespaceByte : NAMESPACE_BYTES) {
```
stIncMale, line 203: I probably completely misunderstand what is going on here, so I apologize for this question if it makes no sense: why cannot we just use `"ns"` instead of `"__"` in `PIPELINE_TEMPLATE`?

src/test/java/com/mongodb/kafka/connect/source/MongoCopyDataManagerTest.java
rozza: Ack will update

src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
rozza: See the related issues in KAFKA-201 for the background information.

src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
rozza: Here I'm going for readability over correctness.  The Java driver is very flexible in parsing json (most json parsers are).  Strictly speaking single quotes are invalid JSON but they are supported by the java driver as well as unquoted strings.

The namespace field was quoted because some special characters need quoting (such as "."). I've updated as its not needed for "__".

src/test/java/com/mongodb/kafka/connect/source/MongoCopyDataManagerTest.java
```diff
@@ -152,34 +149,39 @@ void testAppliesTheExpectedPipelines() {
...
+    Boolean[] hasNextResponses = new Boolean[inputDocs.size()];
     Arrays.fill(hasNextResponses, true);
     hasNextResponses[hasNextResponses.length - 1] = false;
 
     when(cursor.hasNext()).thenReturn(true, hasNextResponses);
     when(cursor.next())
         .thenReturn(
-            docs.get(0),
-            docs.subList(1, docs.size()).toArray(new RawBsonDocument[docs.size() - 1]));
+            inputDocs.get(0),
```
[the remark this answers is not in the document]
rozza: I suppose it allows users to pass an array as the single item returned without it getting expanded via varargs. eg:

```
byte[] bytes = ...;
when(...)
  .thenReturn(bytes);
```

src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
stIncMale: Thank you for explaining this to me!

src/test/java/com/mongodb/kafka/connect/source/MongoCopyDataManagerTest.java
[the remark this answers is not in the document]
stIncMale: Indeed, this is a good point! (though, for this to happen, we need `Byte[]`, not `byte[]`) However, with `thenReturn(T... value)` the same can be achieved by writing
```java
Byte[] bytes = ...;
thenReturn(new Object[] {bytes});
```
If an API user forgets (not unexpectedly) not think about varargs behavior and simply call `thenReturn(bytes)`, it would return `Byte`, not `Byte[]`. If a user does not use the returned value, the user may not notice the problem.

A situation, in which all solutions are flawed. Anyway, thanks you, now I can see why the Mockito may have the API it has.

stIncMale (approved):

src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java
ortex: Hi @rozza 
Could you please give more details?
Maybe make sense to add a comment in the code.

I'm asking in old MR because here was broken backward compatibility in the case of `ns` fields using in `pipeline` config.

P.s: I saw the related issues in KAFKA-201, but it doesn't have much description :(

ortex: Here is example which could be added to `MongoSourceConnectorIntegrationTest`


```java
  @Test
  @DisplayName("Ensure source loads and stream data from certain collections")
  void testSourceLoadsAndStreamDataFromCertainCollections() {
    assumeTrue(isGreaterThanThreeDotSix());
    MongoDatabase db = getDatabaseWithPostfix();
    MongoCollection<Document> coll1 = db.getCollection("coll1");
    MongoCollection<Document> coll2 = db.getCollection("coll2");
    MongoCollection<Document> coll3 = db.getCollection("coll3");

    insertMany(rangeClosed(1, 50), coll1);
    insertMany(rangeClosed(1, 50), coll2);
    insertMany(rangeClosed(1, 50), coll3);

    Properties sourceProperties = new Properties();
    sourceProperties.put(MongoSourceConfig.COPY_EXISTING_CONFIG, "true");
    String namespaceRegex = String.format("(%s\\.coll(1|2))", db.getName());
    sourceProperties.put(MongoSourceConfig.COPY_EXISTING_NAMESPACE_REGEX_CONFIG, namespaceRegex);
    sourceProperties.put(PIPELINE_CONFIG, "[{\"$match\": {\"ns.coll\": {\"$regex\": /^coll(1|2)$/}}}]");
    /* 
    Works with such pipeline
    sourceProperties.put(
        PIPELINE_CONFIG,
        "[{\"$match\": {\"$or\": [{\"ns.coll\": {\"$regex\": /^coll(1|2)$/}}, {\"__.coll\": {\"$regex\": /^coll(1|2)$/}}]}}]");
     */

    addSourceConnector(sourceProperties);

    insertMany(rangeClosed(51, 100), coll1);
    insertMany(rangeClosed(51, 100), coll2);
    insertMany(rangeClosed(51, 100), coll3);

    assertProduced(createInserts(1, 100), coll1);
    assertProduced(createInserts(1, 100), coll2);
    // TODO: assertProducedNone(coll3)
  }
```

ortex: I've created a separate ticket [KAFKA-217](https://jira.mongodb.org/projects/KAFKA/issues/KAFKA-217)

rozza: Thanks @ortex that is indeed an unintended side effect.

PR #18, opened 2020-06-27
Add sinking JSON records from Attunity CDC (Qlik Replicate)

abraham-leal:
Hey guys, recently found myself having to write this custom CDC handler, and wanted to PR it in case it was of interest to integrate it to the project.

This introduces a new handler for the Sink. It largely follows the template of previous RDBMS handlers, but configured to work with the structure and naming of Attunity (now Qlik Repliate) CDC from DBs. It includes some unit testing.

Lmk if interested, if not, we can close this PR.

PR #61, opened 2021-03-11
Qlik Replicate Support

rozza:
Qlik Replicate support.

Changes from #18 - squashed and merged initial commit.

 * Renamed Attunity to Qlik
 * Moved towards the MongoDB CDC organization. 
 * Added Qlik specific OperationType enum.
 * Added support for flattened headers eg `{message: {headers_operation: 'INSERT'}, ...}`
 * Simplified logic determining the CDC operation in `OperationHelper`
 * Refactored tests


rozza:
HI @abraham-leal, this is the cleaned up and refactored version of the qlik support for the 1.5.0 release.  

It would great if you could review.

src/test/java/com/mongodb/kafka/connect/sink/cdc/qlik/rdbms/RdbmsHandlerTest.java
```diff
@@ -0,0 +1,206 @@
...
+import com.mongodb.client.model.DeleteOneModel;
+import com.mongodb.client.model.ReplaceOneModel;
+import com.mongodb.client.model.UpdateOneModel;
+import com.mongodb.client.model.WriteModel;
+
+import com.mongodb.kafka.connect.sink.cdc.qlik.OperationType;
+import com.mongodb.kafka.connect.sink.converter.SinkDocument;
+
+@RunWith(JUnitPlatform.class)
+class RdbmsHandlerTest {
```
abraham-leal, line 50: I see you handle the Qlik envelope in the code, but there are no tests against the case of no qlik envelope.

abraham-leal (approved):
LGTM! Just one comment. Thanks @rozza.

src/test/java/com/mongodb/kafka/connect/sink/cdc/qlik/rdbms/RdbmsHandlerTest.java
rozza: Hi @abraham-leal 

I think all operation parsing should be covered by the OperationHandlerTest. This test just covers a single type of message so not to duplicate the testing.

If I'm not testing something I should could you provide the json of a sample record so I can investigate?

rozza:
Note: Merged on the command line. 

Waiting to see if extra test scenarios are required.

src/test/java/com/mongodb/kafka/connect/sink/cdc/qlik/rdbms/RdbmsHandlerTest.java
abraham-leal: I see only tests for the object created from the OP in `OperationTypeTest`, what I mean is testing for the parsing itself. There can be two types of records by Qlik: With and Without envelope.

With envelope: "{message: { data: {id: 1234, foo: 'bar'}, headers: { operation: 'INSERT'}}}"
Without envelope: "{ data: {id: 1234, foo: 'bar'}, headers: { operation: 'READ'}}"

I see `getSubDocumentOrOriginal` can handle this case, I just don't see it being tested either in `OperationHelper` or `RdbmsHandler`.

rozza: Perfect - we ensure I add tests for that scenario.

Thanks again @abraham-leal

PR #62, opened 2021-03-16
Improved Illegal Operation error messages

rozza:
Provides a relevant error message when users use an illegal pipeline
operation.

KAFKA-174


RWaltersMA (approved):
Since we want to just link to the root Change Stream documentation, let's change the note inl line 413 to

Note: Not all aggregation pipeline operations are suitable for modifying the change stream output.

PR #64, opened 2021-03-19
Ensure deterministic schema record names

rayokota:
This is just an example of how to ensure deterministic schema record names, it needs to be tested thoroughly

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
@@ -76,7 +76,7 @@ public final class BsonDocumentToSchema {
       case TIMESTAMP:
         return Timestamp.builder().optional().build();
       case DOCUMENT:
-        return inferDocumentSchema(bsonValue.asDocument(), true);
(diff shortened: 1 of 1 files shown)


rozza:
Hi @rayokota,

Thanks for the PR. Currently the schema names are deterministic as they use the hashcode of the Schema for the name.

I think the issue is two schemas that could be compatible aren't because they generate different names? That errodes the compatibility. Is that the case?

If an optional field that contained a document from document 1 was missing from document 2. Then the inferred schema from document 2 even using a counter would generate names that aren't compatible.

Ross

rayokota:
Hi @rozza , 

A recent PR to Schema Registry (https://github.com/confluentinc/schema-registry/pull/1775) found that the MongoDB connector is generating incompatible names in some cases, with the hashcodes being different in certain scenarios.  

I think if each `"name"` can just be the field name, that would be ideal.  Then the name at the root level could be something like `"default"`.  WDYT?

cc @rajdangwal

rajdangwal:
Hi @rozza and @rayokota,

Based on our discussions, I have raised another [PR](https://github.com/mongodb/mongo-kafka/pull/65) that uses field names for naming the sub-documents.

Thanks.

rozza:
Thanks @rajdangwal  I'll close this PR and review the new one.

PR #65, opened 2021-03-23
Fix Mongo inferred names for schema records

rajdangwal:
This PR contains changes for using the field names to assign names to sub-documents.
I have tested on some schemas which were reporting backward incompatibility earlier but are now resolved with this fix.
This may require more testing though. Also, the test in src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java needs to be fixed.


rozza:
I've used this as the first commit in #66 - @rajdangwal would you mind taking a quick review?

PR #68, opened 2021-04-05
KAFKA-215: Add override options for error handling properties

C0urante:
Addresses https://jira.mongodb.org/browse/KAFKA-215 by adding new `mongo.`-prefixed variants for error handling configuration properties that either conflict now or may conflict at some point in the future with Connect framework properties, including `errors.tolerance` and `errors.log.enable`. These variants take precedence over their non-prefixed counterparts, but if not specified, the non-prefixed counterpart will be used, preserving backwards compatibility for existing users.


C0urante:
@rozza thoughts?

rozza:
Apologies, for the delay @C0urante I have been on leave.  

Sounds good and I like the option to diverge on a per connector basis. Will look to support and merge in the next release.

C0urante:
Great, thanks @rozza!

rozza (approved):

PR #69, opened 2021-04-05
KAFKA-216: throw when failing to resume change stream for any reason

ncliang:
When the connector is not able to create the cursor to resume the change stream for any reason other than resume token not found, rather than continue silently and not process any data, it should instead fail and allow the user to troubleshoot the issue. The user can decide to resolve the issue in any of the methods that is documented in the log message. Without failing the connector though, it is not obvious to the user that anything has gone wrong with the connector without looking in logs. This is especially problematic with low volume topics that have sporadic traffic patterns in the first place.


rozza:
Hi @ncliang, thanks for the PR.

The usecase here is to be able to support collections / databases that don't exist at the time of configuration.  This makes ops lives easier when setting up new systems.

@RWaltersMA Any thoughts on this issue. See also KAFKA-216

rozza:
I'm going to close this ticket out as it would be a breaking change to existing users. 

I'm happy to look at adding configuration to make it a hard fail - but the default behaviour would have to remain the same, as this is a design decision that we have to keep until a major release.

PR #70, opened 2021-05-27
KAFKA-227: Wrap nullable value returned from WriteModelStrategy…

didiez:
… using Optional.ofNullable(value)

https://jira.mongodb.org/browse/KAFKA-227


rozza:
Many thanks @didiez

PR #73, opened 2021-06-10
Fixed bug in LazyBsonDocument#clone ignoring any changes made once un…

rozza:
…wrapped.

KAFKA-218

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Changelog
 
+## 1.6.0
+
+### Improvements
+
+### Bug Fixes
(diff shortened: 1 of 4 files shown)


rozza:
LazyBsonDocument protects users from having to have valid Json / BsonDocuments as their SinkRecord Key or Value.  This is because a BsonDocument is only needed if the user wants to action the key or value part of the SourceRecord.

The issue with clone is it didn't account for any changes made to the unwrapped BsonDocument - so field renaming then cloning would lose any of the cloned changes.

stIncMale (approved):

PR #74, opened 2021-06-11
ShadowJar updates

rozza:
New Jar: `mongo-kafka-connect-<version>-confluent.jar` contains the connector and mongodb / bson code.
Updated Jar: `mongo-kafka-connect-<version>-all.jar` now also includes the Avro dependency that users not using confluent's runtime will also need.

KAFKA-222


jyemin (approved):
Nice Gradle work.

LGTM

PR #77, opened 2021-07-06
Improved timeseries usability

rozza:
Added conversion tool for non date (String, longs) time data

KAFKA-232


src/main/java/com/mongodb/kafka/connect/sink/processor/TimeseriesTimeFieldAutoConversion.java
```diff
@@ -0,0 +1,92 @@
...
+              if (timeField.isNumber()) {
+                convertedValue =
+                    tryToConvert(() -> new BsonDateTime(timeField.asNumber().longValue()));
+              } else {
+                convertedValue =
+                    tryToConvert(
+                        () -> {
+                          Instant dateTime =
+                              LocalDateTime.parse(timeField.asString().getValue(), formatter)
+                                  .toInstant(ZoneOffset.UTC);
```
stIncMale, line 73: We need to document that configured format is used for parsing strings via `LocalDateTime.parse` because this does have functional implications (or we can document the implications):

1. A parsed string must have not only date information but also time information to be parsed successfully regardless of the format specified.
2. The time offset present in a parsed string is ignored regardless of the format specified.

While the first implication is obvious, the second may not be as obvious, so tested it. Consider the format `uuuu-MM-dd['T'][ ]HH:mm:ss[.][SSSSSS][SSS][X]`, which effectively differs from the default one only by allowing any zone offsets (`[X]`) as opposed to allowing only no or explicit zero offset (`['Z']`):

```java
DateTimeFormatter formatter = DateTimeFormatter
        .ofPattern("uuuu-MM-dd['T'][ ]HH:mm:ss[.][SSSSSS][SSS][X]")
        .withResolverStyle(ResolverStyle.LENIENT); // or STRICT
String text = "1970-01-01T00:00:00.000-0100";
System.out.println(LocalDateTime.parse(text, formatter).toInstant(ZoneOffset.UTC).getEpochSecond());
System.out.println(OffsetDateTime.parse(text, formatter).toInstant().getEpochSecond());
```

This code prints
```
0
3600
```
where 3600 is what a user would expect as a result of parsing, but 0 is what he gets because of us using `LocalDateTime`.

src/integrationTest/java/com/mongodb/kafka/connect/sink/MongoSinkTaskIntegrationTest.java
```diff
@@ -81,6 +81,53 @@ void testSinkProcessesSinkRecords() {
...
+                        doc.put("ts", "2021-07-13T12:00:00.001Z");
+                        break;
+                      case 4:
+                        doc.put("ts", "2021-07-13 12:00:01.001");
+                        break;
+                      case 5:
+                        doc.put("ts", "2021-07-13 12:00:01");
+                        break;
+                      default:
+                        doc.put("ts", now);
```
stIncMale, line 119: If we want to check how `TimeseriesTimeFieldAutoConversion` converts strings for `TIMESERIES_TIMEFIELD_AUTO_CONVERSION_DATE_FORMAT_DEFAULT`, we should create a unit test for this purpose and test various inputs there. An integration test should only check that a sink task successfully creates sink record even if the time field is a string, provided that automatic conversion is enabled.

stIncMale (changes requested):

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -374,7 +375,24 @@ public String value() {
...
           + "time series. Possible values: \"seconds\" \"minutes\" \"hours\".";
-  public static final String TIMESERIES_GRANULARITY_DEFAULT = "";
+
+  public static final String TIMESERIES_TIMEFIELD_AUTO_CONVERSION_CONFIG =
+      "timeseries.timefield.auto.convert";
+  private static final String TIMESERIES_TIMEFIELD_AUTO_CONVERSION_DISPLAY =
+      "Convert the field to a BSON datetime type.";
+  private static final String TIMESERIES_TIMEFIELD_AUTO_CONVERSION_DOC =
+      "Converts the timeseries field to a BSON datetime type. "
+          + "If the value is a numeric value it will use it as the milliseconds from epoch. "
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/kafka/connect/sink/processor/TimeseriesTimeFieldAutoConversion.java
rozza: Updated.

src/integrationTest/java/com/mongodb/kafka/connect/sink/MongoSinkTaskIntegrationTest.java
rozza: Done

src/main/java/com/mongodb/kafka/connect/sink/processor/TimeseriesTimeFieldAutoConversion.java
```diff
@@ -0,0 +1,92 @@
...
+  private final String fieldName;
+  private final DateTimeFormatter formatter;
+
+  TimeseriesTimeFieldAutoConversion(final MongoSinkTopicConfig config) {
+    super(config);
+    fieldName = config.getString(TIMESERIES_TIMEFIELD_CONFIG);
+    formatter =
+        DateTimeFormatter.ofPattern(
+                config.getString(TIMESERIES_TIMEFIELD_AUTO_CONVERSION_DATE_FORMAT_CONFIG))
+            .withResolverStyle(ResolverStyle.LENIENT);
```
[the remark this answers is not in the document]
rozza: I've updated and no longer specify a resolver style. As this is user based we will need to accommodate `yyyy`  because it will be what users are used to.

src/main/java/com/mongodb/kafka/connect/util/SimpleDateTimeParser.java
```diff
@@ -0,0 +1,40 @@
...
+  public SimpleDateTimeParser(final String dateTimePattern) {
+    this.formatter = DateTimeFormatter.ofPattern(dateTimePattern);
+  }
+
+  public long toEpochMilli(final String dateTimeString) {
+    TemporalAccessor parsed = formatter.parse(dateTimeString);
+    if (parsed.isSupported(ChronoField.INSTANT_SECONDS)) {
+      return Instant.from(parsed).toEpochMilli();
+    }
+    return LocalDateTime.from(parsed).atZone(ZoneOffset.UTC).toInstant().toEpochMilli();
```
stIncMale, line 38: I have never had a task of parsing arbitrary strings as time, and this seems like a great approach!

I would like to suggest a few improvements with regard to documenting the behavior (`TIMESERIES_TIMEFIELD_AUTO_CONVERSION_DOC`):

1. A string representation must contain both date and time information in order to be parsed successfully.
2. If a string representation does not contain time-zone offset, then the extracted date and time is interpreted as UTC.

src/test/java/com/mongodb/kafka/connect/util/SimpleDateTimeParserTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+public class SimpleDateTimeParserTest {
+
+  @Test
+  @DisplayName("test default date string")
+  void testDefaultDateStrings() {
+    SimpleDateTimeParser simpleDateTimeParser =
+        new SimpleDateTimeParser("yyyy-MM-dd['T'][ ]HH:mm:ss[.][SSSSSS][SSS][ ][X]");
```
stIncMale, line 30: The pattern here and in `MongoSinkConfigTest` are specified explicitly instead of being specified via the constant `TIMESERIES_TIMEFIELD_AUTO_CONVERSION_DATE_FORMAT_DEFAULT`. If we change the default pattern in the future (if the new pattern broadens the spectrum of strings that fit it, it is a backward-compatible change), neither of the tests will automatically pick up the new default value.

If marking `MongoSinkTopicConfig.TIMESERIES_TIMEFIELD_AUTO_CONVERSION_DATE_FORMAT_DEFAULT` as `public` is not a good idea, maybe we could move the constant somewhere where it can be public and then reuse in other places, including `MongoSinkTopicConfig`? Or maybe it can be just made public in `MongoSinkTopicConfig`.

src/main/java/com/mongodb/kafka/connect/sink/processor/TimeseriesTimeFieldAutoConversion.java
stIncMale: 👍

src/main/java/com/mongodb/kafka/connect/util/SimpleDateTimeParser.java
rozza: Done

src/test/java/com/mongodb/kafka/connect/util/SimpleDateTimeParserTest.java
rozza: Done.

src/main/java/com/mongodb/kafka/connect/util/FlexibleDateTimeParser.java
```diff
@@ -0,0 +1,49 @@
...
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.temporal.ChronoField;
+import java.time.temporal.TemporalAccessor;
+
+public class FlexibleDateTimeParser {
+
+  public static final String DEFAULT_DATE_TIME_FORMATTER_PATTERN =
+      "yyyy-MM-dd['T'][ ][HH:mm:ss[[.][SSSSSS][SSS]][ ]VV[ ]'['VV']'][HH:mm:ss[[.][SSSSSS][SSS][ ]]X][HH:mm:ss[[.][SSSSSS][SSS]]";
```
stIncMale, line 30: This pattern evolved to look similar to a regular expression in its readability :) I noticed that it contains 21 `[` and 20 `]`, which means one `]` is missing (it is unfortunate that `DateTimeFormatter` does not thoroughly check that the patten is well-formed). If we format the pattern for readability, it is easy to spot the missing bracket:

```java
"yyyy-MM-dd['T'][ ]"
+ "["
  + "HH:mm:ss" + "[" + "[.][SSSSSS][SSS]" + "]" + "[ ]VV[ ]'['VV']'"
+ "]"
+ "["
  + "HH:mm:ss" + "[" + "[.][SSSSSS][SSS][ ]" + "]" + "X"
+ "]"
+ "["
  + "HH:mm:ss" + "[" + "[.][SSSSSS][SSS]" + "]"
//']' is missing
;
```

Btw, what do you think about formatting the pattern in the code in a similar (not necessary the same way) to improve readability?

src/test/java/com/mongodb/kafka/connect/util/FlexibleDateTimeParserTest.java
```diff
@@ -0,0 +1,118 @@
...
+            assertThrows(
+                DateTimeParseException.class,
+                () ->
+                    DEFAULT_DATE_TIME_PARSER.toEpochMilli("1970-01-01T02:00:00+01:00[Asia/Paris]"),
+                "Invalid Zone"));
+  }
+
+  @Test
+  @DisplayName("Alternative date time formats")
+  void testAlternativeDateTimeTormats() {
```
stIncMale, line 112: We discovered that the behavior of this test depends on the default locale, e.g., if the corresponding language tag is `en`, then the test passes, but if it is `en-CA`, then the test fails. As a result, Ross proposed to specify the default locale explicitly and allow users to override the explicit default.

stIncMale (changes requested):

src/main/java/com/mongodb/kafka/connect/util/FlexibleDateTimeParser.java
rozza: Great catch and idea, it is a funky pattern :) 

There is auto formatting so I've done the best I can and hope its still readable.

src/test/java/com/mongodb/kafka/connect/util/FlexibleDateTimeParserTest.java
rozza: Great catch as that would have been found by a user and required a patch release :+1:

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -398,6 +399,13 @@ public String value() {
   private static final String TIMESERIES_TIMEFIELD_AUTO_CONVERSION_DATE_FORMAT_DEFAULT =
       DEFAULT_DATE_TIME_FORMATTER_PATTERN;
 
+  public static final String TIMESERIES_TIMEFIELD_AUTO_CONVERSION_LOCALE_LANGUAGE_TAG_CONFIG =
+      "timeseries.timefield.auto.convert.locale.language.tag";
+  private static final String TIMESERIES_TIMEFIELD_AUTO_CONVERSION_LOCALE_LANGUAGE_TAG_DISPLAY =
+      "The DateTimeFormatter locale language tag too use: Defaults to using the neutral Locale.ROOT.";
+  private static final String TIMESERIES_TIMEFIELD_AUTO_CONVERSION_LOCALE_LANGUAGE_TAG_DOC =
+      "The DateTimeFormatter locale language tag too use with the date pattern: Defaults to Locale.ROOT.";
```
stIncMale, line 407: [Nit]
too -> to (the same is applicable to line 405 above)

src/main/java/com/mongodb/kafka/connect/util/FlexibleDateTimeParser.java
```diff
@@ -23,16 +25,30 @@
...
+      format("[%s%s[ ]X]", TIME, OPTIONAL_NANO_SECONDS);
+  private static final String OPTIONAL_TIME = format("[%s%s]", TIME, OPTIONAL_NANO_SECONDS);
+
   public static final String DEFAULT_DATE_TIME_FORMATTER_PATTERN =
-      "yyyy-MM-dd['T'][ ][HH:mm:ss[[.][SSSSSS][SSS]][ ]VV[ ]'['VV']'][HH:mm:ss[[.][SSSSSS][SSS][ ]]X][HH:mm:ss[[.][SSSSSS][SSS]]";
+      "yyyy-MM-dd" // Year month day
+          + "[['T'][ ]]" // optional T or space
+          + OPTIONAL_TIME_WITH_TIMEZONE // optional time with timezone
+          + OPTIONAL_TIME_WITH_OFFSET // optional time without timezone or offset
+          + OPTIONAL_TIME; // optional time without timezone or offset
```
stIncMale, line 45: [Nit]
These comments seem to duplicate constant names, so not sure if they are useful.

Regardless of the above, the comment `OPTIONAL_TIME_WITH_OFFSET // optional time without timezone or offset`  should probably say "with offset", but is currently saying "without timezone or offset".

src/main/java/com/mongodb/kafka/connect/util/FlexibleDateTimeParser.java
```diff
@@ -23,16 +25,30 @@
...
 
 public class FlexibleDateTimeParser {
 
+  private static final String TIME = "HH:mm:ss";
+  private static final String OPTIONAL_NANO_SECONDS = "[[.][SSSSSS][SSS]]";
+  private static final String OPTIONAL_TIME_WITH_TIMEZONE =
+      format("[%s%s[ ]VV[ ]'['VV']']", TIME, OPTIONAL_NANO_SECONDS);
+  private static final String OPTIONAL_TIME_WITH_OFFSET =
+      format("[%s%s[ ]X]", TIME, OPTIONAL_NANO_SECONDS);
+  private static final String OPTIONAL_TIME = format("[%s%s]", TIME, OPTIONAL_NANO_SECONDS);
```
stIncMale, line 38: 👍 for the idea to give names to different pieces of the pattern.

stIncMale (commented):
The changes are good, there are only a few typos that are worth fixing.

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
rozza: Done

src/main/java/com/mongodb/kafka/connect/util/FlexibleDateTimeParser.java
rozza: Done

rozza:
Nit fixed and merged manually

PR #79, opened 2021-07-07
Fixed Source Connector recovery during getMore call

rozza:
KAFKA-230 KAFKA-219

### DIFF
--- a/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
+++ b/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
@@ -25,6 +25,7 @@ import static java.lang.String.format;
 import static java.util.Collections.singletonList;
 import static java.util.Collections.singletonMap;
 import static java.util.stream.Collectors.toList;
+import static java.util.stream.IntStream.range;
 import static java.util.stream.IntStream.rangeClosed;
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -69,6 +70,7 @@ import org.bson.json.JsonWriterSettings;
 
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.model.Updates;
 
 import com.mongodb.kafka.connect.log.LogCapture;
 import com.mongodb.kafka.connect.mongodb.ChangeStreamOperations.ChangeStreamOperation;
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -750,6 +752,66 @@ void testErrorToleranceNoneSupport() {
...
+      coll.replaceOne(
+          new Document("_id", 3),
+          new Document("_id", 3).append("x", new byte[(1024 * 1024 * 16) - 30]));
+      coll.updateOne(
+          new Document("_id", 3),
+          Updates.combine(Updates.unset("x"), Updates.set("y", new byte[(1024 * 1024 * 16) - 30])));
+      insertMany(rangeClosed(4, 5), coll);
+      List<SourceRecord> poll = getNextResults(task);
+
+      assertTrue(poll.size() >= 3 && poll.size() <= 4); // Confirms error loss.
```
stIncMale, line 790: Ross, could you please help me understand the scenario and expectations here?

1. The maximum BSON document size [16 megabytes](https://docs.mongodb.com/manual/reference/limits/#mongodb-limit-BSON-Document-Size), not mebibytes. Why 1024 * 1024 (mebi) is used instead of 1000 (mega)?
2. I assume that `append("x", new byte[(1024 * 1024 * 16) - 30])` and `set("y", new byte[(1024 * 1024 * 16) - 30])` are used to exceed the maximum BSON document size for the corresponding change stream document. And `-30` is used to make sure the document in the collection stays within the limit and operations `replaceOne`, `updateOne` succeed? Is this correct?
3. Why do we need two operations `replaceOne` and `updateOne` that produce change events exceeding the size limit?
4. Why do we expect the source task to produce either 3 or 4 records? Shouldn't it produce 3 records as a result of `insertMany(rangeClosed(1, 3), coll)`, omit the two records corresponding to large change events and produce 2 more records as a result of `insertMany(rangeClosed(4, 5), coll)`?

stIncMale (changes requested):

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -456,12 +485,26 @@ private boolean invalidatedResumeToken(final MongoCommandException e) {
     return e.getErrorCode() == INVALIDATED_RESUME_TOKEN_ERROR;
   }
 
-  private boolean resumeTokenNotFound(final MongoCommandException e) {
-    String errorMessage = e.getErrorMessage().toLowerCase(Locale.ROOT);
-    return errorMessage.contains(RESUME_TOKEN)
+  private boolean checkChangeStreamNotValid(final Exception e) {
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -588,26 +631,37 @@ private void setCachedResultAndResumeToken() {
...
-          }
-          cursor = null;
-        }
+      } catch (MongoException e) {
+        closeCursor();
         if (isRunning.get()) {
           LOGGER.info(
-              "An exception occurred when trying to get the next item from the Change Stream: {}",
-              e.getMessage());
+              "An exception occurred when trying to get the next item from the Change Stream");
```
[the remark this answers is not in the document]
rozza: Done

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
rozza: > 1. The maximum BSON document size 16 megabytes, not mebibytes. Why 1024 * 1024 (mebi) is used instead of 1000 (mega)?

Here it doesn't really matter as either is a valid insert / update.  1024 * 1024 * 16 is approx the max accepted size for an insert document. I think the terms are used synonymously internally.

> 2. I assume that append("x", new byte[(1024 * 1024 * 16) - 30]) and set("y", new byte[(1024 * 1024 * 16) - 30]) are used to exceed the maximum BSON document size for the corresponding change stream document. And -30 is used to make sure the document in the collection stays within the limit and operations replaceOne, updateOne succeed? Is this correct?

The original replace and update are valid size wise. But the update breaks the change stream requirements because `updateLookup` includes the delta between the fields for updates.

 > 3. Why do we need two operations replaceOne and updateOne that produce change events exceeding the size limit?
We probably don't I tried to trigger it with an replaceOne but its the updateOne that breaks the size limit due to delta issue. Note: The [replace event|https://docs.mongodb.com/manual/reference/change-events/#replace-event] doesn't include any deltas.

> 4. Why do we expect the source task to produce either 3 or 4 records? Shouldn't it produce 3 records as a result of insertMany(rangeClosed(1, 3), coll), omit the two records corresponding to large change events and produce 2 more records as a result of insertMany(rangeClosed(4, 5), coll)?

So this came out in the testing, I need to carefully break the change stream so that I get some results then a failure. Turns out adding a replaceOne produces events for some versions of MongoDB (pre 4.4). The updateOne breaks all implementations.  But most importantly, it allows some results through the change stream before breaking it.

On reflection and with further testing, its probably to fragile, so I'll simplify the test.

rozza: Regarding max size this is the error the user sees: `Size must be between 0 and 16793600(16MB)`

From: https://github.com/mongodb/mongo/blob/3b8806bef0f65df11e4d49786bf962ce0dd6b4dd/src/mongo/bson/bsonobj.cpp#L90-L101 so looks like the terms are used synonymously

rozza: Done.

stIncMale: Thank you, Ross! Now the test is much more clear to me.

P.S. 16793600 bits is greater than 16 MiB (and is definitely greater than 16 MB), so 16 MB specified in the error message is an approximation, got it.

stIncMale (approved):

PR #81, opened 2021-07-20
Fix connection validator user privilege check

rozza:
Now uses connection status to determine the users privileges.
Handles authentication mechanisms that don't require a username.

KAFKA-238

### DIFF
--- a/src/integrationTest/java/com/mongodb/kafka/connect/ConnectorValidationIntegrationTest.java
+++ b/src/integrationTest/java/com/mongodb/kafka/connect/ConnectorValidationIntegrationTest.java
@@ -474,7 +474,14 @@ public final class ConnectorValidationIntegrationTest {
   private void assertInvalidSource(final Map<String, String> properties, final String configName) {
     Optional<ConfigValue> configValue =
         getSourceErrors(properties).stream().filter(cv -> cv.name().equals(configName)).findFirst();
-    assertTrue(configValue.isPresent());
+    assertTrue(
+            configValue.isPresent(),
+            format(
+                    "No error for '%s': %s\nErrors: %s\nProperties: %s",
+                    configName,
+                    properties.getOrDefault(configName, ""),
+                    getSourceErrors(properties),
+                    properties));
     assertFalse(
         configValue.get().errorMessages().isEmpty(), format("No error for '%s'", configName));
   }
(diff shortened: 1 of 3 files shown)


rozza:
Changing to draft - the build is failing - investigating

src/integrationTest/java/com/mongodb/kafka/connect/ConnectorValidationIntegrationTest.java
```diff
@@ -474,7 +474,14 @@ private void assertInvalidSource(final Map<String, String> properties) {
   private void assertInvalidSource(final Map<String, String> properties, final String configName) {
     Optional<ConfigValue> configValue =
         getSourceErrors(properties).stream().filter(cv -> cv.name().equals(configName)).findFirst();
-    assertTrue(configValue.isPresent());
+    assertTrue(
```
rozza, line 477: Improve debugging

src/main/java/com/mongodb/kafka/connect/util/ConnectionValidator.java
```diff
@@ -182,26 +168,65 @@ public static void validateUserHasActions(
...
+          .ifPresent(
+              c ->
+                  c.addErrorMessage(
+                      "Invalid user permissions authentication failed. " + e.getMessage()));
     } catch (Exception e) {
       LOGGER.warn("Permission validation failed due to: {}", e.getMessage(), e);
     }
   }
 
+  private static List<String> validateRoleHasActions(
```
stIncMale, line 180: This method behaves similarly to `removeUserActions`: it gets a set of actions and returns a subset representing those actions that are not permitted. However, the naming is different, which negatively contributes to readability.

Another thing I noticed is that the method `removeUserActions` does not modify the passed `actions` collection, which yet its name implies modification.

Can we name this methods in a more consistent and descriptive way?

stIncMale (changes requested):

src/integrationTest/java/com/mongodb/kafka/connect/ConnectorValidationIntegrationTest.java
```diff
@@ -474,7 +474,14 @@ private void assertInvalidSource(final Map<String, String> properties) {
...
+            configValue.isPresent(),
+            format(
+                    "No error for '%s': %s\nErrors: %s\nProperties: %s",
```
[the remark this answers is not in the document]
rozza: `assertInvalidSource` expects there to be an error for the provided configuration.  The errors are filtered so if the configValue isn't present there has been no error.

src/main/java/com/mongodb/kafka/connect/util/ConnectionValidator.java
rozza: Will do

src/integrationTest/java/com/mongodb/kafka/connect/ConnectorValidationIntegrationTest.java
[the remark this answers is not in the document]
stIncMale: Thank you, I have now understood that the `Optional<ConfigValue>` is supposed to contain not a valid `ConfigValue`, but a `ConfigValue` that caused an error.

stIncMale (approved):

PR #84, opened 2021-11-01
Update gradlew to 7.2

rozza:
Added gradle.properties to support spotless / google format java via compile exports

Updated the following to support Java 17:

  * Spotless plugin
  * Spotbugs plugin
  * Buildconfig plugin
  * Junit 5 dependency
  * Mockito dependency
  * Hamcrest dependency

KAFKA-256


build.gradle.kts
```diff
@@ -123,18 +125,18 @@ val gitVersion: String by lazy {
...
+    buildConfigField("String", "NAME", "\"mongo-kafka\"")
+    buildConfigField("String", "VERSION", provider { "\"${gitVersion}\"" })
 }
 
 /*
  * Testing
  */
 sourceSets.create("integrationTest") {
     java.srcDir("src/integrationTest/java")
-    resources.srcDir("src/integrationTest/resources")
```
rozza, line 137: This is already copied and the build errored with a duplicate copy message so removed.

build.gradle.kts
```diff
@@ -32,9 +32,9 @@ plugins {
     `maven-publish`
     signing
     checkstyle
-    id("de.fuerstenau.buildconfig") version "1.1.8"
-    id("com.github.spotbugs") version "4.6.0"
-    id("com.diffplug.spotless") version "5.10.0"
+    id("com.github.gmazzo.buildconfig") version "3.0.3"
```
rozza, line 35: Had to migrate to an alternative plugin that doesn't use removed dependency types (compile)

build.gradle.kts
```diff
@@ -49,7 +49,7 @@ java {
 
 repositories {
     mavenCentral()
-    maven("http://packages.confluent.io/maven/")
+    maven("https://packages.confluent.io/maven/")
```
rozza, line 52: Fixed warning https ftw.

build.gradle.kts
```diff
@@ -85,9 +85,11 @@ dependencies {
     mongoAndAvroDependencies("org.apache.avro:avro:${project.extra["avroVersion"]}")
 
     // Unit Tests
-    testImplementation("org.junit.jupiter:junit-jupiter:${project.extra["junitJupiterVersion"]}")
-    testImplementation("org.junit.platform:junit-platform-runner:${project.extra["junitPlatformVersion"]}")
-    testImplementation("org.hamcrest:hamcrest-junit:${project.extra["hamcrestVersion"]}")
+    testImplementation(platform("org.junit:junit-bom:${project.extra["junitJupiterVersion"]}"))
+    testImplementation("org.junit.jupiter:junit-jupiter")
+    testImplementation("org.junit.platform:junit-platform-runner")
+    testImplementation("org.apiguardian:apiguardian-api:1.1.2") // https://github.com/gradle/gradle/issues/18627
```
rozza, line 91: Gradle is slightly broken - so once its fixed this can be removed.

src/integrationTest/java/com/mongodb/kafka/connect/ConnectorValidationIntegrationTest.java
```diff
@@ -475,13 +475,13 @@ private void assertInvalidSource(final Map<String, String> properties, final Str
...
         getSourceErrors(properties).stream().filter(cv -> cv.name().equals(configName)).findFirst();
     assertTrue(
-            configValue.isPresent(),
-            format(
-                    "No error for '%s': %s\nErrors: %s\nProperties: %s",
-                    configName,
-                    properties.getOrDefault(configName, ""),
-                    getSourceErrors(properties),
-                    properties));
+        configValue.isPresent(),
```
rozza, line 478: spotless auto format change

src/main/java/com/mongodb/kafka/connect/util/ConnectionValidator.java
```diff
@@ -125,8 +125,7 @@ public void clusterDescriptionChanged(
   /**
    * Validates that the user has the required action permissions
    *
-   * <p>Uses the connection status privileges information to check the required action permissions</p>
-   *
+   * <p>Uses the connection status privileges information to check the required action permissions
```
rozza, line 128: spotless auto format changes

jyemin (approved):

build.gradle.kts
```diff
@@ -108,6 +105,13 @@ dependencies {
 
 tasks.withType<JavaCompile> {
     options.encoding = "UTF-8"
+    options.release.set(8)
+}
+
+java {
```
rozza, line 111: Utilising the new toolchain feature

build.gradle.kts
```diff
@@ -123,18 +127,24 @@ val gitVersion: String by lazy {
...
+    buildConfigField("String", "NAME", "\"mongo-kafka\"")
+    buildConfigField("String", "VERSION", provider { "\"${gitVersion}\"" })
 }
 
 /*
  * Testing
  */
+tasks.register<Test>("testsOnJava8") {
+    javaLauncher.set(javaToolchains.launcherFor {
+        languageVersion.set(JavaLanguageVersion.of(8))
```
rozza, line 142: Added a new test task (unit tests) running on Java 8

gradle.properties
```diff
@@ -0,0 +1,3 @@
+org.gradle.java.installations.auto-download=false
```
rozza, line 1: Turned off auto download and setup environmental variables for the jdk versions (I wasn't sure /opt/java would be auto discovered).

gradle.properties
```diff
@@ -0,0 +1,3 @@
...
+org.gradle.java.installations.fromEnv=JDK8,JDK17
+org.gradle.jvmargs=-XX:+HeapDumpOnOutOfMemoryError -Xmx1G --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
```
rozza, line 3: The cause of all the issues - googleformat required all theses exports.

.evergreen/config.yml
```diff
@@ -378,14 +378,6 @@ axes:
         variables:
           TOPOLOGY: "replica_set"
 
-  - id: "jdk"
```
rozza, line 381: JDK removed - as the gradle file now has tasks for running specific jdk versions.

rozza (commented):
@jyemin had to do some work for evergreen and to get spotless to work with googleformat.

build.gradle.kts
rozza: https://docs.gradle.org/current/userguide/toolchains.html

gradle.properties
jyemin: Ack. But did they change?  It looks the same as the deleted line.

.evergreen/run-unit-tests.sh
```diff
@@ -3,18 +3,16 @@
...
 #            Main Program                  #
 ############################################
 
-
-echo "Running tests with ${JDK} connecting to $MONGODB_URI"
+echo "Running unit tests on JDK17 and JDK8"
 
 ./gradlew -version
 ./gradlew --stacktrace --info test
+./gradlew --stacktrace --info testsOnJava8
```
jyemin, line 18: Should we also be running integration tests on Java 8?

build.gradle.kts
```diff
@@ -110,6 +105,13 @@ dependencies {
 
 tasks.withType<JavaCompile> {
     options.encoding = "UTF-8"
+    options.release.set(8)
```
jyemin, line 108: So this replaces the source/target compatibility?

build.gradle.kts
jyemin: How does it find Java 8?

build.gradle.kts
rozza: Correct

.evergreen/run-unit-tests.sh
rozza: I'm not sure we need to, I just want to ensure we can run on Java 8

build.gradle.kts
rozza: Its automatic - but we provide hints via the gradle.property and the JDK8 variables

gradle.properties
rozza: They didnt change - I tried to set them in build.gradle.kts but that failed.

.evergreen/run-unit-tests.sh
jyemin: I just wasn't sure how much unit test coverage there is

build.gradle.kts
jyemin: I'm just trying to wrap my head around this change.  I've always thought the build scripts should be agnostic to JDK version, so it looks weird to me that we have a task like this.  Is there a way to do it more flexibly, perhaps similar to what we do in the driver with jdkHome?  Something like:

```
./gradlew -PwithJDKVersion=8 test
```

and then set the language version based on that property (or leave it unset if the property is not specified).

rozza: There is some coupling as we have to pick a version to compile against, I also provide hints as to where the java versions are installed in via the JDK8/JDK17 variables.

That said I like continuing to use the matrix and we already dynamically set the test jdkHome - so I've used that process but updated it to the toolchain syntax. :+1:

build.gradle.kts
```diff
@@ -165,6 +161,12 @@ tasks.withType<Test> {
         events("passed", "skipped", "failed")
     }
 
+    val javaVersion: Int = (project.findProperty("jdkVersion") as String? ?: defaultJdkVersion.toString()).toInt()
```
rozza, line 164: Allow users to set a custom `jdkVersion` to run the tests against.

rozza:
Re requesting a review - evergreen is now all happy and we pass in the jdkVersion for the tests to use.

jyemin (approved):
I like it.  LGTM.

rozza:
Rebased and merged

PR #86, opened 2021-11-02
[KAFKA-262] Support user defined topic separator

wangxianghu:
Hi
I wrote this PR to make mongo-kafka support user defined topic separator.
currently, mong-kafka use "." to build topics, which might not be allowed in some companies. So it is better to make it configurable.

I introduced topic.separator param in MongoSourceConfig it is "." by default. if the user want use other separator to build topics, just config this param. (E.g “topic.separator”：“-”)

This pr is tested by com.mongodb.kafka.connect.source.topic.mapping.DefaultTopicMapperTest and it runs ok in my local


src/main/java/com/mongodb/kafka/connect/source/topic/mapping/DefaultTopicMapper.java
```diff
@@ -72,7 +74,7 @@ public String getTopic(final BsonDocument changeStreamDocument) {
       return dbName;
     }
     String collName = getStringFromPath(COLL_FIELD_PATH, changeStreamDocument);
-    String namespace = collName.isEmpty() ? dbName : dbName + SEPARATOR + collName;
+    String namespace = collName.isEmpty() ? dbName : dbName + separator + collName;
```
stIncMale: Thank you, @wangxianghu for the proposal and the PR. I made some changes explained below.

We must not use a user-provided separator when constructing the namespace that is used to look up the mapped topic name in `topicNamespaceMap`. The map uses `.` to separate DB names and collection names, and if we use here a user-provided separator, then the map syntax becomes dependent on this separator, which must not happen.

The new commit also refactors the existing `DefaultTopicMapperTest.testProducesTheExpectedTopic` to test not only the implicit default separator, but also the explicitly configured one. This way we don't need to introduce a new test. The original implementation of the support of `TOPIC_SEPARATOR_CONFIG` in `DefaultTopicMapper` fails the test for the reason expressed above.

wangxianghu:
Thanks @stIncMale

PR #89, opened 2021-12-01
Throw ConnectException if resume of change stream is not possible

rajdangwal:
Issue:
One of the MongoDbAtlasSource Connector on Confluent Cloud is seeing the exception `com.mongodb.MongoQueryException: Query failed with error code 286 and error message 'Error on remote shard SOMENAME.mongodb.net:27017 :: caused by :: Resume of change stream was not possible, as the resume point may no longer be in the oplog.'` within the mongo client code but not surfacing it up; causing the connector to run indefinitely without producing any new records.

The exception is thrown here: https://github.com/mongodb/mongo-kafka/blob/28603b3fd78415c616f3363bc79ec947b0b702f0/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java#L635
The corresponding catch block just logs the exception and the connector continues to run.

Mitigating action:
Throw the ConnectException in the catch block when client gets this Exception (error code 286); as the exception says that resume of change stream itself is not possible (this is not a transient error).
Maybe this is something similar to this: https://github.com/mongodb/mongo-kafka/blob/28603b3fd78415c616f3363bc79ec947b0b702f0/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java#L486
The connector should fail in such case instead of just running without producing any records.

Testing:
Haven't tested the changes as I don't have clear path to reproduce the issue. Also, the actual fix may differ.

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
@@ -648,6 +648,10 @@ public final class MongoSourceTask extends SourceTask {
             LOGGER.info(
                 "An exception occurred when trying to get the next item from the Change Stream", e);
           }
+          if (e instanceof MongoCommandException &&
+              ((MongoCommandException) e).getErrorCode() == 286) {
+            throw new ConnectException("Failed to resume change stream", e);
+          }
         }
         return Optional.empty();
       } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


rajdangwal:
@rozza please take a look. Thanks. 🙂

rozza:
Closing, as the durability of the connector is governed by the users error tolerance configuration. Secondly, the MongoDbAtlasSource is a separate codebase managed by the Confluent team.

PR #91, opened 2021-12-02
Create MongoSinkTask unit tests for DQL-related behavior

stIncMale:
This is a subtask of KAFKA-257 created to produce smaller PRs. It is based on the changes done in https://github.com/mongodb/mongo-kafka/pull/88.

KAFKA-268

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTask.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTask.java
@@ -21,21 +21,14 @@ package com.mongodb.kafka.connect.sink;
 import static com.mongodb.kafka.connect.sink.MongoSinkConfig.PROVIDER_CONFIG;
 import static com.mongodb.kafka.connect.util.ConfigHelper.getMongoDriverInformation;
 import static com.mongodb.kafka.connect.util.ServerApiConfig.setServerApi;
-import static com.mongodb.kafka.connect.util.TimeseriesValidation.validateCollection;
 
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.errors.ConnectException;
-import org.apache.kafka.connect.errors.DataException;
 import org.apache.kafka.connect.sink.ErrantRecordReporter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTask;
@@ -43,25 +36,15 @@ import org.apache.kafka.connect.sink.SinkTaskContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.bson.BsonDocument;
-
-import com.mongodb.MongoBulkWriteException;
 import com.mongodb.MongoClientSettings;
-import com.mongodb.MongoException;
-import com.mongodb.MongoNamespace;
-import com.mongodb.bulk.BulkWriteError;
-import com.mongodb.bulk.BulkWriteResult;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoClients;
-import com.mongodb.client.model.BulkWriteOptions;
-import com.mongodb.client.model.WriteModel;
 
 import com.mongodb.kafka.connect.Versions;
 
 public class MongoSinkTask extends SinkTask {
   private static final Logger LOGGER = LoggerFactory.getLogger(MongoSinkTask.class);
   private static final String CONNECTOR_TYPE = "sink";
-  private static final BulkWriteOptions BULK_WRITE_OPTIONS = new BulkWriteOptions();
   private StartedMongoSinkTask startedTask;
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTask.java
```diff
@@ -291,4 +226,91 @@ private String generateWriteErrors(
     }
     return "[" + String.join(", ", errorString) + "]";
   }
+
+  static final class StartedMongoSinkTask {
```
stIncMale, line 230: The refactoring in `MongoSinkTask` did not change any logic, but merely extracted the instance logic that is not related to the Kafka Connect-specific lifecycle of the `MongoSinkTask` to a new class `StartedMongoSinkTask`, which can now be easily unit tested.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
```diff
@@ -0,0 +1,389 @@
...
+                // batch1
+                Records.simpleValid(TEST_TOPIC, 0),
+                // batch2
+                Records.simpleValid(TEST_TOPIC2, 1),
+                Records.simpleValid(TEST_TOPIC2, 2),
+                Records.simpleValid(TEST_TOPIC2, 3),
+                // batch3
+                Records.simpleValid(TEST_TOPIC, 4)),
+            asList(0, 1, 2, 3, 4),
+            emptyList());
```
stIncMale, line 188: KAFKA-257 will change this empty expectation to `asList(2, 3)`:
- 2 will be reported because it has a corresponding write error;
- 3 will be reported because the bulk write is ordered, which means that 3 is not attempted to be written.

rozza (approved):
Looks good - some stylistic comments for consideration but not necessarily requiring action.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    static SinkRecord simpleInvalid(final String topicName, final int idx) {
+      return new SinkRecord(
+          topicName, 0, Schema.STRING_SCHEMA, "", Schema.STRING_SCHEMA, "notJSON_" + idx, idx);
+    }
+
+    static boolean match(
+        final SinkRecord expected, final WriteModel<? extends BsonDocument> actual) {
+      ReplaceOneModel<BsonDocument> writeModel = cast(actual);
```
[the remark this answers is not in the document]
stIncMale: I added `assertTrue(actual instanceof ReplaceOneModel)` to make the expectation clear. If for some reason in the future this method encounters a write model different from `ReplaceOneModel`, I want it to fail as early as possible (it will need to be updated to support comparing that write model), rather than returning `false` and causing a failure in a different place.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
```diff
@@ -0,0 +1,432 @@
...
+  private static final MongoNamespace DEFAULT_NAMESPACE =
+      new MongoNamespace(TEST_DATABASE, "myColl");
+
+  private Map<String, String> properties;
+  private FakeClient client;
+  private FakeErrorReporter errorReporter;
+
+  @BeforeEach
+  void setUp() {
+    properties = new HashMap<>();
```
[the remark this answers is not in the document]
stIncMale: - Renamed `FakeClient` -> `BulkWritesCapturingClient`, `FakeErrorReporter` -> `InMemoryErrorReporter`.

- I don't think having these classes in `SinkTestHelper` is useful at this point. If we introduce other test classes that need either the same of very similar functionality, then extracting fake implementations for reusing may become useful.

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTask.java
stIncMale: I did not make it a top-level class initially, because I was trying to minimize the changes in `MongoSinkTask`. But I am happy to have such a change. Done.

> So this is just a mechanism for setting the mongo client?

Yes it's just manual dependency injection (DI):

- It results in dependencies of `StartedMongoSinkTask` (sink config, MongoDB client, error reporter) being expressed clearly, which improves readability;
- It allows the code that creates such tasks to control the implementation of dependencies, which decouples `StartedMongoSinkTask` from those implementations. This is crucial for unit tests, because they now can easily specify mocked implementations.

> the stages are all still linked to the Kafka lifestyle - eg: stop and put

The important part is that the two aforementioned points are still true. I poorly expressed the idea in the initial comment, but hopefully now it is clearer.

> Is there a cleaner way to achieve that for testing.

I don't think there is. DI is applied here right to its point. And of all different styles of DI, constructor injection is the cleanest one.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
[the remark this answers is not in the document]
rozza: Ack

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
rozza: Ack

rozza (approved):
LGTM!

PR #93, opened 2021-12-13
KAFKA 271

rozza:
Update gradle and test with JDK 17


build.gradle.kts
```diff
@@ -155,6 +161,12 @@ tasks.withType<Test> {
         events("passed", "skipped", "failed")
     }
 
+    val javaVersion: Int = (project.findProperty("jdkVersion") as String? ?: defaultJdkVersion.toString()).toInt()
```
stIncMale, line 164: Where does the `jdkVersion` project property may come from when we are building? I can't find it anywhere in the codebase, and it does not seem to be mentioned in Gradle docs. Is it possible that `javaVersion` was supposed to be used here? As this is the system property specified by scripts that run tests, e.g., by `run-integration-tests.sh`:
```shell
./gradlew -PjavaVersion=$JAVA_VERSION
```

rozza: Thanks @stIncMale - I did the same originally with the java driver :disappointed: and failed to update this copy pasta.

stIncMale (approved):

PR #95, opened 2021-12-18
Report MongoDB errors to the dead letter queue (DLQ)

stIncMale:
The corresponding documentation changes needed are explained in [DOCSP-20141](https://jira.mongodb.org/browse/DOCSP-20141).

KAFKA-257

### DIFF
--- a/src/integrationTest/java/com/mongodb/kafka/connect/sink/MongoSinkTaskIntegrationTest.java
+++ b/src/integrationTest/java/com/mongodb/kafka/connect/sink/MongoSinkTaskIntegrationTest.java
@@ -42,6 +42,8 @@ import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTask;
 import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
+import org.apache.log4j.spi.LoggingEvent;
+import org.apache.log4j.spi.ThrowableInformation;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
@@ -143,11 +145,16 @@ public class MongoSinkTaskIntegrationTest extends MongoKafkaTestCase {
           logCapture.getEvents().stream()
               .filter(e -> e.getLevel().equals(Level.ERROR))
               .anyMatch(
-                  e ->
-                      e.getMessage()
-                          .toString()
-                          .startsWith(
-                              "WriteErrors: [BulkWriteError{writeModel=InsertOneModel{document={\"_id\": 4}}")));
+                  e -> {
+                    String message = String.valueOf(e.getMessage());
+                    return message.contains("Failed to put into the sink the following records")
+                        && message.contains("{\"_id\": 4}");
+                  }),
+          () ->
+              "Captured"
+                  + logCapture.getEvents().stream()
+                      .map(MongoSinkTaskIntegrationTest::toString)
+                      .collect(toList()));
     }
   }
 
@@ -541,5 +548,16 @@ public class MongoSinkTaskIntegrationTest extends MongoKafkaTestCase {
                     valueSupplier.apply(i),
                     kafkaOffsetSupplier.apply(i).longValue()))
         .collect(toList());
+  }
+
+  private static String toString(final LoggingEvent e) {
+    ThrowableInformation eInfo = e.getThrowableInformation();
+    Throwable t = eInfo == null ? null : eInfo.getThrowable();
+    return LoggingEvent.class.getSimpleName()
+        + "{message="
+        + e.getMessage()
+        + ", exception="
+        + t
+        + '}';
   }
 }
--- a/src/main/java/com/mongodb/kafka/connect/sink/NoStackTraceDlqException.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/NoStackTraceDlqException.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mongodb.kafka.connect.sink;
+
+import org.apache.kafka.connect.errors.DataException;
+import org.apache.kafka.connect.sink.ErrantRecordReporter;
+
+import com.mongodb.lang.Nullable;
+
+/**
+ * An exception that is meant to be reported to the DLQ via {@link ErrantRecordReporter}. The
+ * {@linkplain Class#getName() class name} and {@linkplain #getMessage() message} must be treated as
+ * public API. This is necessary to allow the possibility of processing the DLQ in an automated way.
+ * The length of the {@linkplain #getStackTrace() stack trace} is always 0, which means that this
+ * exception is suitable for reporting only situations with the stack trace being inconsequential.
+ */
+abstract class NoStackTraceDlqException extends DataException {
+  private static final long serialVersionUID = 1L;
+
+  NoStackTraceDlqException(@Nullable final String message) {
+    super(message);
+  }
+
+  @Override
+  public final Throwable fillInStackTrace() {
+    return this;
+  }
+}
--- a/src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
@@ -17,18 +17,11 @@
  */
 package com.mongodb.kafka.connect.sink;
 
-import static com.mongodb.kafka.connect.util.Assertions.assertNotNull;
-import static com.mongodb.kafka.connect.util.Assertions.assertTrue;
 import static com.mongodb.kafka.connect.util.TimeseriesValidation.validateCollection;
-import static java.util.Collections.singletonList;
 
-import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
@@ -42,15 +35,14 @@ import org.bson.BsonDocument;
 
 import com.mongodb.MongoBulkWriteException;
 import com.mongodb.MongoNamespace;
-import com.mongodb.bulk.BulkWriteError;
 import com.mongodb.bulk.BulkWriteResult;
-import com.mongodb.bulk.WriteConcernError;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.model.BulkWriteOptions;
 import com.mongodb.client.model.WriteModel;
-import com.mongodb.lang.Nullable;
 
-final class StartedMongoSinkTask {
+import com.mongodb.kafka.connect.sink.dlq.AnalyzedBatchFailedWithBulkWriteException;
+
+public final class StartedMongoSinkTask {
   private static final Logger LOGGER = LoggerFactory.getLogger(MongoSinkTask.class);
   private static final BulkWriteOptions BULK_WRITE_OPTIONS = new BulkWriteOptions();
 
@@ -114,7 +106,13 @@ final class StartedMongoSinkTask {
       Thread.currentThread().interrupt();
       throw new DataException("Rate limiting was interrupted", e);
     } catch (RuntimeException e) {
-      handleTolerableWriteException(batch, e, config.logErrors(), config.tolerateErrors());
+      handleTolerableWriteException(
+          batch.stream()
+              .map(MongoProcessedSinkRecordData::getSinkRecord)
+              .collect(Collectors.toList()),
+          e,
+          config.logErrors(),
+          config.tolerateErrors());
     }
   }
 
@@ -143,17 +141,14 @@ final class StartedMongoSinkTask {
   }
 
   private void handleTolerableWriteException(
-      final List<MongoProcessedSinkRecordData> batch,
+      final List<SinkRecord> batch,
       final RuntimeException e,
       final boolean logErrors,
       final boolean tolerateErrors) {
-    List<SinkRecord> records =
-        batch.stream()
-            .map(MongoProcessedSinkRecordData::getSinkRecord)
-            .collect(Collectors.toList());
     if (e instanceof MongoBulkWriteException) {
       AnalyzedBatchFailedWithBulkWriteException analyzedBatch =
-          new AnalyzedBatchFailedWithBulkWriteException(records, (MongoBulkWriteException) e);
+          new AnalyzedBatchFailedWithBulkWriteException(
+              batch, (MongoBulkWriteException) e, errorReporter);
       if (logErrors) {
         analyzedBatch.log();
       }
@@ -164,7 +159,7 @@ final class StartedMongoSinkTask {
       }
     } else {
       if (logErrors) {
-        log(records, e);
+        log(batch, e);
       }
       if (!tolerateErrors) {
         throw new DataException(e);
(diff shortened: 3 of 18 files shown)


src/main/java/com/mongodb/kafka/connect/util/Assertions.java
```diff
@@ -0,0 +1,102 @@
...
+ * All {@code assert...} and {@code fail} methods throw {@link AssertionError} and should be used to
+ * check conditions which may be violated if and only if the driver code is incorrect. The intended
+ * usage of these methods is the same as of the <a
+ * href="https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html">Java {@code
+ * assert} statement</a>. The reason for not using the {@code assert} statements is that they are
+ * not always enabled. We prefer having internal checks always done at the cost of our code doing a
+ * relatively small amount of additional work in production. The {@code assert...} methods return
+ * values to open possibilities of being used fluently.
+ */
+public final class Assertions {
```
stIncMale, line 30: This class was copied from the Java driver repository (it's not public API there, so we can't just reuse it).

src/integrationTest/java/com/mongodb/kafka/connect/sink/MongoSinkTaskIntegrationTest.java
```diff
@@ -542,4 +549,15 @@ public void stop() {
                     kafkaOffsetSupplier.apply(i).longValue()))
         .collect(toList());
   }
+
+  private static String toString(final LoggingEvent e) {
```
rozza, line 553: Nice - much easier to debug :)

src/main/java/com/mongodb/kafka/connect/sink/NoStackTraceDlqException.java
```diff
@@ -0,0 +1,41 @@
...
+ * An exception that is meant to be reported to the DLQ via {@link ErrantRecordReporter}. The
+ * {@linkplain Class#getName() class name} and {@linkplain #getMessage() message} must be treated as
+ * public API. This is necessary to allow the possibility of processing the DLQ in an automated way.
+ * The length of the {@linkplain #getStackTrace() stack trace} is always 0, which means that this
+ * exception is suitable for reporting only situations with the stack trace being inconsequential.
+ */
+abstract class NoStackTraceDlqException extends DataException {
+  private static final long serialVersionUID = 1L;
+
+  NoStackTraceDlqException(@Nullable final String message) {
```
rozza, line 33: Consider moving these exceptions into their own package.

src/main/java/com/mongodb/kafka/connect/sink/NoStackTraceDlqException.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * An exception that is meant to be reported to the DLQ via {@link ErrantRecordReporter}. The
+ * {@linkplain Class#getName() class name} and {@linkplain #getMessage() message} must be treated as
+ * public API. This is necessary to allow the possibility of processing the DLQ in an automated way.
+ * The length of the {@linkplain #getStackTrace() stack trace} is always 0, which means that this
+ * exception is suitable for reporting only situations with the stack trace being inconsequential.
+ */
+abstract class NoStackTraceDlqException extends DataException {
+  private static final long serialVersionUID = 1L;
```
rozza, line 31: Suggestion add an abstract version field - So all versions of NoStackTraceDlqExceptions have to implement their own version number

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -103,22 +110,11 @@ private void bulkWriteBatch(final List<MongoProcessedSinkRecordData> batch) {
...
       throw new DataException("Rate limiting was interrupted", e);
-    } catch (Exception e) {
-      if (config.logErrors()) {
-        LOGGER.error("Failed to write mongodb documents", e);
-      }
-      if (!config.tolerateErrors()) {
-        throw new DataException("Failed to write mongodb documents", e);
-      }
+    } catch (RuntimeException e) {
+      handleTolerableWriteException(batch, e, config.logErrors(), config.tolerateErrors());
```
rozza, line 117: Nit: Its seems a little odd to use batch here but unpack it straight away in the method to a list of SinkRecords but unpack the configs here.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -146,48 +142,129 @@ private static void checkRateLimit(final MongoSinkTopicConfig config)
...
-                + writeModels.get(bulkWriteError.getIndex())
-                + ", code="
-                + bulkWriteError.getCode()
-                + ", message='"
-                + bulkWriteError.getMessage()
-                + '\''
-                + ", details="
-                + bulkWriteError.getDetails()
-                + '}');
+  private class AnalyzedBatchFailedWithBulkWriteException {
```
rozza, line 175: Consider moving into its own file - only uses the error reporter which can be passed in and the static logger.

rozza: Perhaps add a comment as to the purpose of this class

src/main/java/com/mongodb/kafka/connect/sink/WriteConcernException.java
```diff
@@ -0,0 +1,42 @@
...
+ * (marked with {@code v}) will be incremented.
+ */
+final class WriteConcernException extends NoStackTraceDlqException {
+  private static final long serialVersionUID = 1L;
+
+  WriteConcernException(final WriteConcernError error) {
+    super(
+        String.format(
+            Locale.ENGLISH,
+            "v=1, code=%d, codeName=%s, message=%s, details=%s",
```
rozza, line 36: Suggestion set v1 to be static of the class.

src/main/java/com/mongodb/kafka/connect/sink/WriteSkippedException.java
```diff
@@ -0,0 +1,37 @@
...
+ * This exception signifies that writing a {@link SinkRecord} was not attempted by a MongoDB server,
+ * e.g., because writes where done via an {@linkplain BulkWriteOptions#isOrdered() ordered}
+ * {@linkplain MongoCollection#bulkWrite(List, BulkWriteOptions) bulk write}, and the server failed
+ * to write a preceding {@link SinkRecord}.
+ */
+final class WriteSkippedException extends NoStackTraceDlqException {
+  private static final long serialVersionUID = 1L;
+
+  WriteSkippedException() {
+    super(null);
```
rozza, line 35: Suggestion: Put a user friendly message here, explaining it was not attempted by the server.

src/main/java/com/mongodb/kafka/connect/util/Assertions.java
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * @param value A value to check.
+   * @param <T> The type of {@code value}.
+   * @return {@code null}.
+   * @throws AssertionError If {@code value} is not {@code null}.
+   */
+  @Nullable
+  public static <T> T assertNull(@Nullable final T value) throws AssertionError {
```
rozza, line 38: Suggestion - remove unused assertions and add then when / if needed.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
```diff
@@ -240,18 +256,40 @@ static boolean match(
...
-  private static final class InMemoryErrorReporter
-      implements Consumer<MongoProcessedSinkRecordData> {
-    private final List<MongoProcessedSinkRecordData> reported = new ArrayList<>();
+  private static final class InMemoryErrorReporter implements BiConsumer<SinkRecord, Exception> {
+    private final List<ReportedData> reported = new ArrayList<>();
 
     @Override
-    public void accept(final MongoProcessedSinkRecordData mongoProcessedSinkRecordData) {
-      reported.add(mongoProcessedSinkRecordData);
+    public void accept(final SinkRecord record, final Exception e) {
```
rozza, line 263: `MongoProcessedSinkRecordData` already supports a record and an exception - could / should it be used?

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
```diff
@@ -240,18 +256,40 @@ static boolean match(
...
+      reported.add(new ReportedData(record, e));
     }
 
-    List<MongoProcessedSinkRecordData> reported() {
+    List<ReportedData> reported() {
       return unmodifiableList(reported);
     }
+
+    static final class ReportedData {
```
rozza, line 271: This looks familiar to the `MongoProcessedSinkRecordData` - could we use a single data structure?

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -146,48 +142,129 @@ private static void checkRateLimit(final MongoSinkTopicConfig config)
...
+        StartedMongoSinkTask.log(
+            recordsFailedWithWriteConcernError.values(), assertNotNull(writeConcernException));
+      }
+      if (!skippedRecords.isEmpty()) {
+        StartedMongoSinkTask.log(skippedRecords.values(), writeSkippedException);
+      }
+      LOGGER.error(null, e);
+    }
+
+    /**
```
rozza, line 239: Nit: Spotless rearranges this javadoc (or it does for me locally)

rozza (changes requested):
Like it, couple of minor comments are:

Can the exceptions be moved into their own package?
`StartedMongoSinkTask` is quite large can the `AnalyzedBatchFailedWithBulkWriteException` be moved into its own class (it doesnt share much).

Finally, we use a similar data structure - SinkRecord and Exception in both processing the sink records and via a BiConsumer - Do you think we should use a single data structure? We also use one in testing.

src/main/java/com/mongodb/kafka/connect/sink/NoStackTraceDlqException.java
stIncMale: `serialVersionUID` must be a `static` field, there is no way to express it as an `abstract` method, because its value is a characteristic of a class, not that of an object. Could you please clarify the suggestion?

P.S. Note also that `ErrantRecordReporter.report` does not use the Java serialization mechanism when reporting exceptions.

src/main/java/com/mongodb/kafka/connect/sink/WriteConcernException.java
stIncMale: I thought about it before submitting the PR, and it appears that specifying it as part of the format string literal is better because otherwise the format will look like `"v=%d, ..."`, and it is likely that the docs team would specify it that way in the official docs, and we would have to clarify in the doc ticket that the format should be documented as `"v=1, ..."`. It's a minor negative thing, but since I did not find any positive things regarding using the `"v=%d, ..."` format in the code, I decided to have the value of the version as part of the format literal.

What benefits do you see with regard to using the `"v=%d, ..."` format in the code, and having the value version as a `static` field?

src/main/java/com/mongodb/kafka/connect/sink/WriteSkippedException.java
stIncMale: I did not specify any message here to spare us from sending tons of such strings to the DLQ, and to spare the DLQ from storing them. Note that if a huge ordered batch fails with a single write exception in the beginning of the batch, then `WriteSkippedException` is reported for most of its records. Reporting only static (compile-time) data does not seem reasonable here, and that's why this exception does not have any message. Users will see the name of the exception in the DLQ, and have the explanation of what the exception means in the official docs.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
stIncMale: `MongoProcessedSinkRecordData.exception` represents an exception that happens during [post-processing](https://docs.mongodb.com/kafka-connector/current/sink-connector/fundamentals/post-processors/). Refactoring the code in a way that uses this field to express exceptions that happen when writing to MongoDB is not only inconvenient (try to imagine how you would change the `StartedMongoSinkTask.bulkWriteBatch` to pass bulk write exceptions via `MongoProcessedSinkRecordData.exception`), but also violates the semantics of the `MongoProcessedSinkRecordData.exception`.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
stIncMale: `ReportedData` represents a pair of arguments passed to `StartedMongoSinkTask.errorReporter`. This pair of arguments is not the same thing as the pair `(MongoProcessedSinkRecordData.sinkRecord, MongoProcessedSinkRecordData.exception)`:
- It is semantically a different thing, as I mentioned [above](https://github.com/mongodb/mongo-kafka/pull/95#discussion_r779258628).
- `MongoProcessedSinkRecordData` does not provide a way to create an instance with an arbitrary exception. Refactoring this class to allow this does not make sense from the standpoint of the purpose of this class.

src/main/java/com/mongodb/kafka/connect/sink/NoStackTraceDlqException.java
stIncMale: Done.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
stIncMale: Done, moved to the `com.mongodb.kafka.connect.sink.dlq` package together with DQL-related exceptions.

src/main/java/com/mongodb/kafka/connect/util/Assertions.java
stIncMale: Done.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
stIncMale: It rearranges the doc for me also, but somehow it did not rearrange it previously (I run `./gradlew check -x test -x integrationTest` before committing). Anyway, done.

src/main/java/com/mongodb/kafka/connect/sink/WriteConcernException.java
rozza: Only benefit was to make it a forcing function to ensure that a version is set, also a variable is more self documenting that it being hidden in the string. The end result to a user of course would be the same. Its more to help with understanding the code in 6 months or for a new developer to understand what it means.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
rozza: Ack

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
rozza: :+1: thanks :)

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
rozza: I can live with that - although processed is an open term processed by what? the connector, mongodb?  

Had this work been done at the same time I'd probably have gone for `PostProcessedSinkRecordData` as a more explicit name for the class.

src/main/java/com/mongodb/kafka/connect/sink/WriteSkippedException.java
rozza: Ack, good point :+1:

src/main/java/com/mongodb/kafka/connect/sink/NoStackTraceDlqException.java
rozza: Apologies - crossed wires here I was asking to consider adding a version field for each DLQ Exception. We've discussed this else where - so resolving this.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -164,107 +159,19 @@ private void handleTolerableWriteException(
...
-        }
-      }
-    }
+  public static void log(final Collection<SinkRecord> records, final RuntimeException e) {
+    LOGGER.error("Failed to put into the sink the following records: {}", records, e);
   }
 
-  private static void log(final Collection<SinkRecord> records, final RuntimeException e) {
-    LOGGER.error("Failed to put into the sink the following records: {}", records, e);
+  public static void log(final RuntimeException e) {
```
rozza, line 174: I think this can be removed?

src/main/java/com/mongodb/kafka/connect/sink/dlq/AnalyzedBatchFailedWithBulkWriteException.java
```diff
@@ -0,0 +1,140 @@
...
+  public void log() {
+    if (!recordsFailedWithWriteError.isEmpty()) {
+      recordsFailedWithWriteError.forEach(
+          (i, recordFailedWithWriteError) ->
+              StartedMongoSinkTask.log(
+                  singletonList(recordFailedWithWriteError.getKey()),
+                  recordFailedWithWriteError.getValue()));
+    }
+    if (!recordsFailedWithWriteConcernError.isEmpty()) {
+      StartedMongoSinkTask.log(
```
rozza, line 104: Suggestion: decouple the logging mechanism from StartedMongoSinkTask  - pass in a ` BiConsumer<Collection<SinkRecord>, RuntimeException>` to handle the logging.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -164,107 +159,19 @@ private void handleTolerableWriteException(
...
-              recordFailedWithWriteConcernError, assertNotNull(writeConcernException));
-          continue;
-        }
-        SinkRecord skippedRecord = skippedRecords.get(i);
-        if (skippedRecord != null) {
-          errorReporter.accept(skippedRecord, writeSkippedException);
-        }
-      }
-    }
+  public static void log(final Collection<SinkRecord> records, final RuntimeException e) {
```
rozza, line 170: If you opt to decouple the logging mechanism and inject it instead this could be:

```
private static BiConsumer<Collection<SinkRecord>, RuntimeException>  ERROR_LOGGER =
          (records, e) -> LOGGER.error("Failed to put into the sink the following records: {}", records, e);```

rozza (commented):
Looks really good, I have a couple of minor suggestions inject in the logger to `AnalyzedBatchFailedWithBulkWriteException`. If you did that I'd also support changing the signature of the `errorreporter` to match exception types.

src/main/java/com/mongodb/kafka/connect/sink/WriteConcernException.java
stIncMale: Done. Extracted it to a static field.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
stIncMale: Sorry, for this rudiment. Done.

src/main/java/com/mongodb/kafka/connect/sink/dlq/AnalyzedBatchFailedWithBulkWriteException.java
stIncMale: Done.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
stIncMale: I decoupled `AnalyzedBatchFailedWithBulkWriteException` from `StartedMongoSinkTask`, but replacing this method with an object does not seem beneficial for the code nonetheless. Take a look, maybe you'll agree.

stIncMale:
> If you did that I'd also support changing the signature of the errorreporter to match exception types.

I replaced `BiConsumer<SinkRecord, Exception>` with `ErrantRecordReporter`. This way the interface is less generic, which makes the code clearer.

src/test/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTaskTest.java
stIncMale: Ross, do you want me to change the class name from `MongoProcessedSinkRecordData` to `PostProcessedSinkRecordData` in this PR?

src/main/java/com/mongodb/kafka/connect/util/Assertions.java
rozza: Ack

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
rozza: Ack

rozza (approved):
Thanks for your patience!  LGTM :)

PR #97, opened 2022-01-12
Added `copy.existing.allow.disk.use` configuration

rozza:
Allows the copy existing aggregation to use temporary
disk storage if required. Defaults to true but can be
disabled if the user doesn't have the permissions
for disk access.

KAFKA-265

### DIFF
--- a/src/test/java/com/mongodb/kafka/connect/source/MongoCopyDataManagerTest.java
+++ b/src/test/java/com/mongodb/kafka/connect/source/MongoCopyDataManagerTest.java
@@ -17,6 +17,7 @@ package com.mongodb.kafka.connect.source;
 
 import static com.mongodb.kafka.connect.source.MongoCopyDataManager.ALT_NAMESPACE_FIELD;
 import static com.mongodb.kafka.connect.source.MongoSourceConfig.COLLECTION_CONFIG;
+import static com.mongodb.kafka.connect.source.MongoSourceConfig.COPY_EXISTING_ALLOW_DISK_USE_CONFIG;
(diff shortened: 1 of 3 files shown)


src/test/java/com/mongodb/kafka/connect/source/MongoCopyDataManagerTest.java
```diff
@@ -138,6 +140,7 @@ void testAppliesTheExpectedPipelines() {
     when(mongoDatabase.getCollection(TEST_COLLECTION, RawBsonDocument.class))
         .thenReturn(mongoCollection);
     when(mongoCollection.aggregate(expectedPipeline)).thenReturn(aggregateIterable);
+    when(aggregateIterable.allowDiskUse(true)).thenReturn(aggregateIterable);
```
stIncMale, line 143: [optional]

Since here and in all other similar places we don't actually care whether `true` or `false` is passed, let's use `ArgumentMatchers.anyBoolean` instead of `true`.

Alternatively, if we do want to mock only if the default (`true`) value is specified (this does not seem to be the case), then I propose to use `COPY_EXISTING_ALLOW_DISK_USE_DEFAULT` (has to be made package-access) instead of `true`.

stIncMale (approved):
There is only one optional suggestion. Since it is optional, I am approving the PR in case you decide to merge.

src/test/java/com/mongodb/kafka/connect/source/MongoCopyDataManagerTest.java
rozza: Done

rozza:
Manually merged and added to the CHANGELOG.md

PR #99, opened 2022-01-26
fixed source setup in docker example

gasabr:
Added json key and value converter to the MongoDB Kafka
Source Connector, the default Avro converter does not
handle mongo jsons properly adding binary values in front
of each key and value. The behaviour could be verified
by running ./run.sh from the docker folder

KAFKA-285
https://jira.mongodb.org/browse/KAFKA-285


rozza:
Thanks @gasabr for the fix :)

PR #101, opened 2022-02-02
Log incompatible configuration properties

stIncMale:
Specifically, when the CDC handler is configured
(https://docs.mongodb.com/kafka-connector/current/sink-connector/configuration-properties/cdc/)
configuration properties for post processors
https://docs.mongodb.com/kafka-connector/current/sink-connector/configuration-properties/post-processors/
and configuration properties for the ID strategy
are ignored.

KAFKA-277

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkConfig.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkConfig.java
@@ -18,7 +18,31 @@
 
 package com.mongodb.kafka.connect.sink;
 
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.CHANGE_DATA_CAPTURE_HANDLER_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.CHANGE_DATA_CAPTURE_HANDLER_DEFAULT;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DELETE_ON_NULL_VALUES_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DELETE_ON_NULL_VALUES_DEFAULT;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DOCUMENT_ID_STRATEGY_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DOCUMENT_ID_STRATEGY_OVERWRITE_EXISTING_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DOCUMENT_ID_STRATEGY_OVERWRITE_EXISTING_DEFAULT;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DOCUMENT_ID_STRATEGY_UUID_FORMAT_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.FIELD_RENAMER_MAPPING_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.FIELD_RENAMER_MAPPING_DEFAULT;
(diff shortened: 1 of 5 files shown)


src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -597,7 +596,6 @@ public NamespaceMapper getNamespaceMapper() {
     String prefix = format("%s%s.", TOPIC_OVERRIDE_PREFIX, topic);
     List<String> topicOverrides =
         props.keySet().stream().filter(k -> k.startsWith(prefix)).collect(Collectors.toList());
-    logObsoleteProperties(topicOverrides);
```
stIncMale, line 600: I realized that `MongoSinkTopicConfig.validateAll` is called from `ConfigDef.validateAll` (see `MongoSinkConfig.createConfigDef`) with the same `props` as those passed to `ConfigDef.validateAll`. I.e., this call results in duplicate logging.

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkConfig.java
```diff
@@ -18,7 +18,31 @@
...
 package com.mongodb.kafka.connect.sink;
 
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.CHANGE_DATA_CAPTURE_HANDLER_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.CHANGE_DATA_CAPTURE_HANDLER_DEFAULT;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DELETE_ON_NULL_VALUES_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DELETE_ON_NULL_VALUES_DEFAULT;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DOCUMENT_ID_STRATEGY_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DOCUMENT_ID_STRATEGY_OVERWRITE_EXISTING_CONFIG;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DOCUMENT_ID_STRATEGY_OVERWRITE_EXISTING_DEFAULT;
+import static com.mongodb.kafka.connect.sink.MongoSinkTopicConfig.DOCUMENT_ID_STRATEGY_UUID_FORMAT_CONFIG;
```
rozza, line 28: I think the topic configuration should be responsible for logging all topic based configuration.  The top level sink configuration should delegate to each topic configuration.

rozza (changes requested):

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkConfig.java
rozza: The benefits are the topic based configuration manages its own validation and not the higher level sink config.  It seems to be recoupling the design by pushing this upwards.

stIncMale: > the topic based configuration manages its own validation

I don't think this is the case: the validation is currently done (even before this PR) by the static code. I.e., there is no object that represents topic-specific configuration that has an instance method validating it. Instead, there is a static method `MongoSinkTopicConfig.validateAll` doing this, and this static method is called from `ConfigDef.validateAll` defined in `MongoSinkConfig`.

In other words, the proposal will result only in restructuring the static code. No responsibility will be moved from from an object to a different object, because validation is done by static code, not instance methods. I.e, it is simply impossible to say that the topic based configuration will manage its own validation, because the object representing that configuration is created after the static code does the validation.

stIncMale: We discussed this over Zoom, and agreed to move the logic into a separate class. It was moved to a new `SinkConfigSoftValidator` class, what exposes methods `logObsoleteProperties` and `logIncompatibleProperties`, used from `ConfigDef.validateAll`.

rozza (approved):
LGTM

PR #102, opened 2022-02-11
Update the driver version to 4.5.0

stIncMale:
KAFKA-279


rozza (changes requested):
~Looks good but 4.5.0 hasn't been released yet - so release will have to wait until then Or we can fix to a 4.4.x release~

Ignore that - I see 4.5.0 was released :) LGTM!

rozza (approved):

PR #104, opened 2022-02-15
Stop converting unsupported `fixed` Avro types

stIncMale:
We don't support the `FIXED` type (`AvroSchema.validateJsonSchema` fails when when encounters one), but `AvroSchema.createSchema` was written in such a way that could mislead a reader into thinking that we support this type.

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/schema/AvroSchema.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/schema/AvroSchema.java
@@ -134,7 +134,7 @@ public final class AvroSchema {
 
   public static Schema fromJson(final String jsonSchema) {
     org.apache.avro.Schema parsedSchema = validateJsonSchema(jsonSchema);
-    return createSchema(parsedSchema);
+    return createSchema(parsedSchema, false, null, new Context());
   }
 
   static org.apache.avro.Schema parseSchema(final String jsonSchema) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/mongodb/kafka/connect/source/schema/AvroSchema.java
```diff
@@ -145,15 +145,6 @@ public static Schema fromJson(final String jsonSchema) {
...
 
-  static Schema createSchema(final org.apache.avro.Schema avroSchema) {
-    return createSchema(avroSchema, false, new Context());
-  }
-
-  static Schema createSchema(
-      final org.apache.avro.Schema avroSchema, final boolean isOptional, final Context context) {
-    return createSchema(avroSchema, isOptional, null, context);
-  }
-
```
stIncMale, line 156: Removed these methods because they were making code less straightforward without adding something useful. If you think they actually were useful, I will roll this change back and make them `private`.

src/main/java/com/mongodb/kafka/connect/source/schema/AvroSchema.java
```diff
@@ -218,11 +208,12 @@ static Schema createSchema(
                 .filter(s -> s.getType() != org.apache.avro.Schema.Type.NULL)
                 .findFirst();
         if (optionalSchema.isPresent()) {
-          return createSchema(optionalSchema.get(), true, context);
+          return createSchema(optionalSchema.get(), true, null, context);
         }
         throw new IllegalStateException();
       case NULL:
       case ENUM:
+      case FIXED:
```
stIncMale, line 216: `validateJsonSchema` complains when encounters `FIXED`, so there is no reason to support converting `FIXED` in `createSchema`. We can also see that `NULL` and `ENUM` have been treated in the same way as `FIXED`.

rozza (approved):
LGTM

PR #105, opened 2022-05-02
Allow Source Connector to send tombstones on delete events when publishing only fullDocument

cosimomeli:
Hi,
we had the need to send tombstones on delete events to allow consumers to delete data when origin data in deleted.
Currently the Source connector filters out every null fullDocument.
I added a configuration `tombstone.on.delete` to opt-in the feature.
When both `tombstone.on.delete`  and `publish.full.document.only` are true, the delete event with null fullDocument will be sent as tombstone (null value and null schema) to Kafka.
An integrationTest has been created to check for null values.

### DIFF
--- a/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
+++ b/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
@@ -575,6 +575,59 @@ public class MongoSourceTaskIntegrationTest extends MongoKafkaTestCase {
   }
 
   @Test
+  @DisplayName("Test null values are emitted when documents are deleted")
+  void testSourceEmitsNullValuesOnDelete() {
+    try (AutoCloseableSourceTask task = createSourceTask()) {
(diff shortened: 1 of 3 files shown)


cojam-dutchie (approved):
I'm tossing my symbolic approval on this. I'm working with this connector as well and just realized this behavior wasn't supported. I'm producing to a compacted topics using the documentKey as the message key. I would like the connector to produce tombstone records so that disc space will be freed up by the compaction process.

denisw:
This feature is essential for maintaining a compacted Kafka topic that mirrors the contents of a MongoDB collection using `publish.full.document.only`. In addition to space savings, tombstones are also important for data privacy reasons if the documents contain private data.

rozza (approved):
LGTM

rozza:
Had to manually rebase and merged See: 759966ec4eca869bf640678b3be6bd1899bdc07f

**Note:** I renamed the configuration option to: `publish.full.document.only.tombstone.on.delete` to be super explicit and to follow the convention in relating the configuration to the higher level config (`publish.full.document.only`).

Many thanks @cosimomeli for the PR and all those that commented on the ticket.

PR #108, opened 2022-07-14
Support setting a custom deletewritemode.strategy

ArneKlein:
Hi,

we currently have the scenario where we have to create multi version of the same document based on a key in the value and delete all versions based on the key once we receive a tombstone event.

To support this scenario, it would be great to enable setting a custom "deletewritemodel.strategy" just as we currently can for the normal writemodel.strategy

KAFKA-320

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
@@ -175,9 +175,9 @@ public class MongoSinkTopicConfig extends AbstractConfig {
   public static final String DELETE_WRITEMODEL_STRATEGY_CONFIG = "delete.writemodel.strategy";
   private static final String DELETE_WRITEMODEL_STRATEGY_DISPLAY = "The delete writeModel strategy";
   private static final String DELETE_WRITEMODEL_STRATEGY_DOC =
-          "The class the handles how to build the delete write models for the sink documents.";
+      "The class the handles how to build the delete write models for the sink documents.";
   static final String DELETE_WRITEMODEL_STRATEGY_DEFAULT =
-          "com.mongodb.kafka.connect.sink.writemodel.strategy.DeleteOneDefaultStrategy";
+      "com.mongodb.kafka.connect.sink.writemodel.strategy.DeleteOneDefaultStrategy";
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -172,6 +172,13 @@ public String value() {
   static final String WRITEMODEL_STRATEGY_DEFAULT =
       "com.mongodb.kafka.connect.sink.writemodel.strategy.DefaultWriteModelStrategy";
 
+  public static final String DELETEWRITEMODEL_STRATEGY_CONFIG = "deletewritemodel.strategy";
```
rozza, line 175: Rename to `DELETE_WRITEMODEL_STRATEGY_CONFIG` and config to: `delete.writemodel.strategy`

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -508,12 +515,7 @@ PostProcessors getPostProcessors() {
 
   public WriteModelStrategy getWriteModelStrategy() {
```
rozza, line 516: Seems strange to have two methods with the same name. But having different effects (one caching). I'd be tempted to refactor / rename eg: `getWriteModelStrategyFromConfig`

rozza (changes requested):
Apologies for the delay in reviewing. 

I've added a couple of minor nits for being included into master. 
I would like to see some tests added to test the new logic

ArneKlein:
> Apologies for the delay in reviewing.
> 
> I've added a couple of minor nits for being included into master. I would like to see some tests added to test the new logic

Thank you for the review, I added basically the same tests as for getWriteModelStrategy(). Please let me know if this is sufficient

rozza (approved):

rozza:
LGTM

ArneKlein:
Any chance that this can get merged?

jyemin:
Note: the latest patch doesn't compile:

```
com/mongodb/kafka/connect/sink/MongoSinkConfigTest.java:812: error: incompatible types: Optional<WriteModelStrategy> cannot be converted to WriteModelStrategy
               cfg.getMongoSinkTopicConfig(TEST_TOPIC).getDeleteWriteModelStrategy();
```

PR #109, opened 2022-07-20
Add JMX statistics for timings and counts

katcharov:
KAFKA-304

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
@@ -24,6 +24,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import org.apache.kafka.connect.errors.DataException;
@@ -49,11 +50,14 @@ import com.mongodb.kafka.connect.util.jmx.Timer;
 public final class StartedMongoSinkTask {
   private static final Logger LOGGER = LoggerFactory.getLogger(MongoSinkTask.class);
 
+  private static final AtomicInteger NEXT_ID = new AtomicInteger();
+
   private final MongoSinkConfig sinkConfig;
   private final MongoClient mongoClient;
   private final ErrorReporter errorReporter;
   private final Set<MongoNamespace> checkedTimeseriesNamespaces;
 
+  private int id = NEXT_ID.getAndAdd(1);
   private final SinkTaskStatistics statistics;
   private Timer lastTaskInvocation = null;
 
@@ -68,9 +72,8 @@ public final class StartedMongoSinkTask {
     statistics = MBeanServerUtils.registerMBean(new SinkTaskStatistics(), getMBeanName());
   }
 
-  private static String getMBeanName() {
-    // TODO name
-    return "com.mongodb:type=MongoDBKafkaConnector,name=SinkTask" + Thread.currentThread().getId();
+  private String getMBeanName() {
+    return "com.mongodb:type=MongoDBKafkaConnector,name=SinkTask" + id;
   }
 
   /** @see MongoSinkTask#stop() */
--- a/src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
@@ -42,14 +42,20 @@ import com.mongodb.client.model.WriteModel;
 
 import com.mongodb.kafka.connect.sink.dlq.AnalyzedBatchFailedWithBulkWriteException;
 import com.mongodb.kafka.connect.sink.dlq.ErrorReporter;
+import com.mongodb.kafka.connect.util.jmx.MBeanServerUtils;
+import com.mongodb.kafka.connect.util.jmx.SinkTaskStatistics;
+import com.mongodb.kafka.connect.util.jmx.Timer;
 
 public final class StartedMongoSinkTask {
-  private static final Logger LOGGER = LoggerFactory.getLogger(MongoSinkTask.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(StartedMongoSinkTask.class);
 
   private final MongoSinkConfig sinkConfig;
   private final MongoClient mongoClient;
   private final ErrorReporter errorReporter;
   private final Set<MongoNamespace> checkedTimeseriesNamespaces;
+
+  private final SinkTaskStatistics statistics;
+  private Timer lastTaskInvocation = null;
 
   StartedMongoSinkTask(
       final MongoSinkConfig sinkConfig,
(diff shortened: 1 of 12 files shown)


src/main/java/com/mongodb/kafka/connect/util/jmx/SourceTaskStatisticsMBean.java
```diff
@@ -0,0 +1,81 @@
...
+package com.mongodb.kafka.connect.util.jmx;
+
+/**
+ * <p>A standard MBean interface for a {@link com.mongodb.kafka.connect.source.MongoSourceTask}.</p>
+ *
+ * <p>This interface is NOT part of the public API.  Be prepared for non-binary compatible changes in minor releases.</p>
+ *
+ * @since TODO
+ */
+public interface SourceTaskStatisticsMBean {
```
jyemin, line 26: We should distinguish in some way between the two phases of the source task: the copy phase and the change stream phase.  These stats will be misleading if you don't know what phase you're in.  As one example, getMore latencies of 5 seconds are normal in the change stream phase, but highly unusual in the copy phase.  If you can't tell the difference, it's going to be confusing to customers and support engineers.

src/main/java/com/mongodb/kafka/connect/util/jmx/SourceTaskStatisticsMBean.java
```diff
@@ -0,0 +1,81 @@
...
+
+  /** @return Milliseconds spent in the task, including any sub-phases. */
```
jyemin, line 28: All three of these times could (and probably should) be represented as well, e.g.

* taskPollTimeMsLessThanX/Y/Z/... in powers of 10
* readTimeMsLessThanX/Y/Z...
* taskBetweenPollMsLessThanX/Y/Z/... 

and consider renaming these to emphasize that they are totals

jyemin (commented):
Looking good.  Some initial thoughts before I leave for the day.

src/main/java/com/mongodb/kafka/connect/util/jmx/SourceTaskStatistics.java
```diff
@@ -0,0 +1,143 @@
...
+ */
+
+package com.mongodb.kafka.connect.util.jmx;
+
+import java.util.concurrent.TimeUnit;
+
+public class SourceTaskStatistics implements SourceTaskStatisticsMBean {
+
+  // Timings
+  private long taskTimeNanos;
```
jyemin, line 24: I'm not sure what is considered best practice for JMX, but it occurs to me that reads of all of these values will be happening from a different thread than the writes, so some synchronization seems to be required.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -42,15 +42,21 @@
 
 import com.mongodb.kafka.connect.sink.dlq.AnalyzedBatchFailedWithBulkWriteException;
 import com.mongodb.kafka.connect.sink.dlq.ErrorReporter;
+import com.mongodb.kafka.connect.util.jmx.MBeanServerUtils;
+import com.mongodb.kafka.connect.util.jmx.SinkTaskStatistics;
+import com.mongodb.kafka.connect.util.jmx.Timer;
 
 public final class StartedMongoSinkTask {
-  private static final Logger LOGGER = LoggerFactory.getLogger(MongoSinkTask.class);
```
rozza, line 47: IIRC this was left to keep the log class the same - users may already have preferences setup specifically for the MongoSinkTask

rozza (commented):
Looking good - just one note to add regarding logger name.

Also, for future information - The Jira git plugin seems to pick up any commits on any branch, so you don't have to push your development branch (stats) into the mongo repo, you can just PR from your own clone.

src/main/java/com/mongodb/kafka/connect/util/jmx/SourceTaskStatistics.java
katcharov: I guess there's no harm in making these `AtomicLong`. (Though I can't think of a case where this would have a notable effect. JMX doesn't query the properties atomically, so there's already no guarantee that they would correspond to each other at a queried point in time.)

jyemin: This is one of those times when I miss Valentin. :)

I think there are a couple of reasons to use AtomicLong here.  

One is that reads of longs are not guaranteed by the JVM to be atomic.  They can be sheared (not sure that's the right term) so that you can read a value that never actually existed 

The second is that the JVM does not guarantee that an observing thread will ever see updates to these values without some sort of synchronization.  They are allowed to remain in per-core caches indefinitely.

src/main/java/com/mongodb/kafka/connect/util/jmx/SourceTaskStatisticsMBean.java
katcharov: Agreed. It seems there are also some stats (total number of records sent) the user might prefer not to split. We could:

1. Create a separate property for each stat.
2. When the task switches from "copy" to "change stream", switch to a second MBean.
3. Switch the MBean as in #2, but also offer a third MBean where both phases are summed together for convenience.
4. Keep it as is, but introduce a new stat (like "records left to copy") that communicates when the copy phase is over, and assume that the user will notice this stat.

We were leaning towards #3.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -737,4 +797,13 @@ BsonDocument getResumeToken(final MongoSourceConfig sourceConfig) {
     }
     return resumeToken;
   }
+
+  @Override
+  public void commitRecord(SourceRecord record, RecordMetadata metadata) {
```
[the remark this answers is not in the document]
katcharov: This is invoked by Kafka when the record sent to the topic has been acknowledged by the server. The docs for the method specify:

> Commit an individual SourceRecord when the callback from the producer client is received. This method is also called when a record is filtered by a transformation, and thus will never be ACK'd by a broker. In this case metadata will be null.

src/main/java/com/mongodb/kafka/connect/util/jmx/SourceTaskStatisticsMBean.java
jyemin: Defer to subsequent PR

src/main/java/com/mongodb/kafka/connect/util/jmx/SourceTaskStatisticsMBean.java
```diff
@@ -0,0 +1,82 @@
...
+  long getPollTaskTimeMs();
+
+  /** @return Milliseconds spent reading from MongoDB. Included in {@link #getPollTaskTimeMs()}. */
+  long getPollTaskReadTimeMs();
```
jyemin, line 33: Let's separate out getMore here as well.

jyemin (changes requested):

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -185,26 +208,87 @@ public void start(final Map<String, String> props) {
...
+            .addCommandListener(
+                new CommandListener() {
+                  @Override
+                  public void commandSucceeded(final CommandSucceededEvent event) {
+                    String commandName = event.getCommandName();
+                    long elapsedTime = event.getElapsedTime(TimeUnit.NANOSECONDS);
+                    if ("getMore".equals(commandName)) {
+                      currentStatistics.successfulGetMoreCommand();
+                      currentStatistics.getMoreCommandElapsedTimeNanos(elapsedTime);
+                    } else {
```
jyemin, line 222: I don't expect any other commands (expect maybe killCursors?), but just to be a little defensive I would add a check here and below for "find" or "aggregate".

jyemin (changes requested):

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -59,21 +69,40 @@ public final class StartedMongoSinkTask {
     this.mongoClient = mongoClient;
     this.errorReporter = errorReporter;
     checkedTimeseriesNamespaces = new HashSet<>();
+    statistics = MBeanServerUtils.registerMBean(new SinkTaskStatistics(), getMBeanName());
+  }
+
+  private String getMBeanName() {
+    return "com.mongodb:type=MongoDBKafkaConnector,name=SinkTask" + id;
```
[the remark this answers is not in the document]
rozza: Investigating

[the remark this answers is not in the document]
rozza: The [kafka documentation](https://kafka.apache.org/documentation/#common_node_monitoring) seems to have a strange MBean name and I'm wondering if there is some sort of convention based approach.

eg:
```
kafka.streams:type=stream-task-metrics,thread-id=([-.\w]+),task-id=([-.\w]+)
```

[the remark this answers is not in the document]
rozza: I can't see how to access the task id - unfortunately the task context string has it in eg: `WorkerSinkTaskContext{id=MongoSinkConnector-0}` but I can't see how accessible it is.  I would reach out to the apache mailing list or confluent via Robert W and see if there is a better way to access.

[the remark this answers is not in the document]
rozza: https://cwiki.apache.org/confluence/display/KAFKA/KIP-803
https://lists.apache.org/thread/o4f6qbxkl97zytg9znb2dmp867sl7hdg

Looks like its not possible but it is possible to pass an identifier via the `MongoSinkConnector#taskConfigs`  and use the index there for the task id. I think it would correlate to the actual task id.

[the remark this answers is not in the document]
katcharov: I'll defer this. Passing in an index via the config map doesn't seem substantively different. Both approaches would, in normal cases, increment from 0 to N and match up with Kafka's id assignment.

The id is [embedded in the thread name](https://github.com/a0x8o/kafka/blob/master/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerTask.java#L243). I'll try to pull the number out of there, even though it is internal, and fall back to the id if this fails. This seems like the best way to ensure that we match the Kafka JMX names.

PR #110, opened 2022-07-25
start change stream at operation time

addios:
This feature to set change stream to start at specific time. When synchronizing the data from source to destination, it will have a gap of time during exporting and importing data. Some data could be lost between that gap. 
Thus, we can use this feature to start the change stream before exporting data from the source.


rozza:
Hi @addios,

Thanks for the PR. This is a feature we are keen to add for the next release and opted for a slightly different approach. Please see: #125 for more details.

Ross

PR #112, opened 2022-08-03
Add size, lag, and limit metrics

katcharov:
Uses a dynamic MBean to help manage the additional metrics. Some of the metrics have been renamed to better fit as prefixes for the new metrics. The MBean naming is now more consistent with Kafka:

Mongo: `com.mongodb.kafka.connect:type=sink-task-metrics,task=sink-task-0`
Kafka: `kafka.connect:type=sink-task-metrics,connector=MongoSinkConnector,task=0`

The "0" is taken from the current thread, which Kafka names (internally) according to the task number. The stats are logged at the debug level at the end of each task invocation.

Search for `.sample(` in the code changes to see where the numbers come from. The integration tests enumerate the metrics.

https://jira.mongodb.org/browse/KAFKA-64


src/main/java/com/mongodb/kafka/connect/util/jmx/internal/AdditiveMetric.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  public void emit(final Consumer<MetricValue> consumer) {
+    consumer.accept(new MetricValue(name, () -> this.count.get()));
+    for (int i = 0; i < bins.length; i++) {
+      int finalI = i;
+      consumer.accept(
+          new MetricValue(name + "-over-" + limits[i] + unit, () -> bins[finalI].get()));
+    }
+    consumer.accept(new MetricValue(name + "-total-" + unit, () -> this.total.get()));
```
katcharov, line 58: This produces:
```
"task-invocations",
"task-invocations-over-1ms",
"task-invocations-over-10ms",
"task-invocations-over-100ms",
"task-invocations-over-1000ms",
"task-invocations-over-10000ms",
"task-invocations-total-ms",
```
We could also add:
```
"task-invocations-max-ms"
"task-invocations-latest-ms"
```
(These would be added to all 11 "additive" metrics.)

jyemin (commented):
I didn't remember what a DynamicMBean was, and when I looked it up I found something to discuss.  Let's resolve that thread before proceeding.

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MongoMBean.java
```diff
@@ -0,0 +1,142 @@
...
+        list.add(metricsMap.get(name).get());
+      }
+    }
+    return list;
+  }
+
+  @Override
+  public MBeanInfo getMBeanInfo() {
+    List<MBeanAttributeInfo> attrs = new ArrayList<>();
+    for (Metric metric : this.metrics) {
```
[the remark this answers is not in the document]
jyemin: What's weird is that the DynamicMBean Javadoc doesn't state this requirement.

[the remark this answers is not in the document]
katcharov: In the current code, the MBean never changes. Each metric is registered with the MBean on field initialization, before the MBean is registered (and therefore visible for an `getMBeanInfo` invocation). This registration pattern is similar to [what Kafka does](https://github.com/a0x8o/kafka/blob/master/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/ConnectMetricsRegistry.java#L385), and allows fields to be specified as follows:
```
  private Metric failedRecords = registerTotal("records-failed");
  private Metric latestOffsetMs = registerLatest("latest-offset-ms");
  private Metric taskInvocations = registerMs("task-invocations");
```
It is true that we might later decide to register a new metric after the MBean has been constructed, and inadvertently violate this constraint, but this seems improbable, and difficult to prevent in code regardless.

Our tests already assert that each MBean produces a complete set of all metric attributes, after the test completes. I can add a test that confirms that all metric attributes available immediately after the MBean constructor is invoked match that complete set of metric attributes. I think this test would be more explicit about the constraint, and would cover everything we cannot cover in code. Would this address your concern?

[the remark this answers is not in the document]
jyemin: sgtm

jyemin (changes requested):
A few initial comments, but before I go further, I would like to see unit tests written for logic contained in the util (and sub) packages.  It seems easy to do, and would help catch any logic errors that might be missed in code review.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -207,51 +205,68 @@ public void start(final Map<String, String> props) {
     partitionMap = null;
     createPartitionMap(sourceConfig);
 
+    CommandListener statisticsCommandListener =
```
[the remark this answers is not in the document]
katcharov: It seemed more appropriate to move part of the logic into `ResumeTokenUtils` as you suggested below, and the rest of the logic into a couple of methods on MongoSourceTask, similar to `commitRecord(`. Let me know if that doesn't address your concern. (Note that the remaining logic depends on `currentStatistics`, which is a non-final field of MongoSourceTask).

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -207,51 +205,68 @@ public void start(final Map<String, String> props) {
...
+                .map(BsonValue::asDocument)
+                .map(v -> v.get("postBatchResumeToken"))
+                .map(BsonValue::asDocument)
+                .ifPresent(
+                    resumeToken -> {
+                      long resumeTime =
+                          getTimestampFromResumeToken(resumeToken).asTimestamp().getTime();
+                      long opTime = response1.get("operationTime").asTimestamp().getTime();
+                      long offset = resumeTime - opTime;
+                      currentStatistics.getLatestOffsetSecs().sample(offset);
```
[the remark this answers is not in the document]
katcharov: Good catch.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -207,51 +205,68 @@ public void start(final Map<String, String> props) {
...
+              currentStatistics.getSuccessfulInitiatingCommands().sample(elapsedTimeMs);
+            }
+            trackLagStatistics(event.getResponse());
+          }
+
+          private void trackLagStatistics(final BsonDocument response1) {
+            // operationTime, clusterTime (no)
+            // changestreambatchtoken.resumecursor
+
+            Optional.ofNullable(response1)
```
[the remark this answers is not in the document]
katcharov: Fair point. Refactored.

katcharov:
I added tests for the `MongoMBean`. These tests cover the behaviours of the individual metrics classes, and integration with the `MBeanServer`. I also added metrics checks, where appropriate, to the existing integration tests covering the source and sink tasks. In contrast to a mocked unit test of the tasks, this is less coupled to the implementation, and adds additional checks to those existing integration tests. I believe this overlap sufficiently covers the MBean functionality.

In those integration tests, I could not reliably get the counts to match in local testing. For example, the number of successful records was coming in under 100. There is a nice "eventually" mechanism (retries the checks) that I hoped to use for these metrics, but it was periodically failing even after substantially increasing the timeout to 50 seconds. I eventually weakened the tests by comparing against 0. On evergreen, there is also an error that is being reliably picked up in a locally-passing test that I do not expect to see an error in. I would consider these failures outside of the scope of the MBean functionality, though worth investigating.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -267,24 +249,32 @@ public void commandFailed(final CommandFailedEvent event) {
...
     if (sourceRecords != null) {
-      currentStatistics.returnedRecords(sourceRecords.size());
+      currentStatistics.getRecordsReturned().sample(sourceRecords.size());
     }
-    if (!isCopying.get()) {
-      currentStatistics = streamStatistics;
+    currentStatistics.getTaskInvocations().sample(taskTime.getElapsedTime(TimeUnit.MILLISECONDS));
+    lastTaskInvocation = Timer.start();
+
+    if (LOGGER.isDebugEnabled()) {
```
jyemin, line 274: I think it would be better to only print the current statistic.  When we're in the copy state, stream statistics are empty.  And once we're in stream state, copy statistics no long change (I think).

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -828,9 +824,38 @@ BsonDocument getResumeToken(final MongoSourceConfig sourceConfig) {
...
     if (metadata == null) {
-      currentStatistics.filteredRecords(1);
+      currentStatistics.getRecordsFiltered().sample(1);
     } else {
-      currentStatistics.successfulRecords(1);
+      currentStatistics.getRecordsAcknowledged().sample(1);
+    }
+  }
+
+  private void mongoCommandSucceeded(final CommandSucceededEvent event) {
```
jyemin, line 833: I still think that the entire CommandListener implementation should be broken out into a separate class and unit tested, but I can see the argument against, as it's tightly coupled to MongoSourceTask.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -828,9 +824,38 @@ BsonDocument getResumeToken(final MongoSourceConfig sourceConfig) {
...
+    ResumeTokenUtils.getResponseOffsetSecs(event.getResponse())
+        .ifPresent(offset -> currentStatistics.getLatestOffsetSecs().sample(offset));
+  }
+
+  private void mongoCommandFailed(final CommandFailedEvent event) {
+    Throwable e = event.getThrowable();
+    if (e instanceof MongoCommandException) {
+      if (doesNotSupportsStartAfter((MongoCommandException) e)) {
+        // silently ignore this expected exception, which is used to set this.supportsStartAfter
+        return;
```
jyemin, line 850: It could be misleading to ignore these events entirely.  Did you consider whether to treat them as successful commands instead?

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -85,24 +86,44 @@ void stop() {
...
+  }
+
+  private void trackLatestRecordTimestampOffset(final Collection<SinkRecord> records) {
+    OptionalLong latestRecord =
+        records.stream()
+            .filter(v -> v.timestamp() != null)
+            .mapToLong(ConnectRecord::timestamp)
+            .max();
+    if (latestRecord.isPresent()) {
+      long offsetMs = System.currentTimeMillis() - latestRecord.getAsLong();
```
jyemin, line 124: I guess there is nothing we can do about the non-monotonicity of currentTimeMillis, if we want to track this stat?  Should we at least check for negative values here and ignore them?

jyemin (changes requested):

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
katcharov: Negative values seem fine, because if one clock is behind the other, we might start with a stable `-10000`, and perhaps set an alert if it moves up by 5000ms, signalling a lag spike. If we capped the negatives, any change would be obscured.

I imagine that in cases where non-monotonicity comes into play, such as a fast/slow clock being reset, we would want to see the reset reflected here.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
katcharov: It seems that ideally, we might hit the server during initialization to see if it supports `startsAfter`. That wouldn't be as efficient. But if we did that, I would expect such a request to not be counted as a normal message, so on that basis I don't count it here. I agree that this might be confusing for users who enable server logging and are trying to match up message counts. But it seems more confusing to log this as a success when it would show up as an error in such logs. And counting it as an error when it is expected might needlessly worry users. Ignoring seems like the least-bad option, though I don't feel particularly strongly about this.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
katcharov: Why would it be useful to separate copy and stream in JMX, but not in logging?

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
```diff
@@ -0,0 +1,120 @@
...
+      if (server.isRegistered(objectName)) {
+        server.unregisterMBean(objectName);
+      }
+    } catch (Exception e) {
+      // JMX might not be available
+      LOGGER.warn("Unable to unregister MBean " + mBeanName, e);
+    }
+  }
+
+  public static Map<String, Map<String, Long>> getMBeanAttributes(
```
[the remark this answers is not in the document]
katcharov: This is used only by integration tests. If you prefer, I could move this code to `integrationTest`.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
jyemin: Separating is still useful.  I think the analogy would be if we were unregister the copy mbean once the copying phase is over.  And maybe that's a good idea too.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
jyemin: I actually have no idea where the timestamp in ConnectRecord comes from.  Could it be from a different process or server?

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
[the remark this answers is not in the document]
jyemin: oh, yeah, if it's only used by tests let's move it.

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
katcharov: The timestamp type associated with the timestamp is either the "create time" or the "log append" time. I don't know what it could be if not a reasonable approximation of the time the record was put into Kafka (and there's no other field we might use instead). [This SO answer and comment](https://stackoverflow.com/a/62936145) suggest that the default is "local epoch time of the producing system".

jyemin: Yeah, then not a lot we can do here.  I'll resolve

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
katcharov: I changed it to log only the current.

I don't think we should unregister the copy mbean, because it might be sampled, and then accrue some counts, and then be unregistered before it is sampled again. (This is incidentally not an issue with logging, since no counts are accrued between logging and the switch from copy to stream.)

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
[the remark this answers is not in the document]
katcharov: I moved it, and then moved it back, since I needed it for a unit test in `test`, in addition to the usage in `integrationTest`. I think the integration test coverage is sufficient in this case.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
katcharov: All stat invocations are now explicitly covered, including the getmore failure. I added stats checks to existing integration tests. The only one that wasn't already covered was `records-filtered` (a Kafka callback stat).

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
```diff
@@ -0,0 +1,126 @@
...
+
+public final class MBeanServerUtils {
+  private static final Logger LOGGER = LoggerFactory.getLogger(MBeanServerUtils.class);
+  private static final AtomicInteger NEXT_ID = new AtomicInteger();
+
+  private MBeanServerUtils() {
+    // util class
+  }
+
+  public static String registerMBean(final DynamicMBean mBean, final String mBeanName) {
```
rozza, line 52: Nit: Annotate nullable?

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
```diff
@@ -0,0 +1,126 @@
...
+        | IntrospectionException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Thread names have the form "task-thread-MongoSinkConnector-0", where the number suffix is the
+   * id shown in Kafka's JMX MBean.
+   */
+  public static String taskIdFromCurrentThread() {
```
rozza, line 121: Nice work around the lack of official way to get the id.

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/AdditiveMetric.java
```diff
@@ -0,0 +1,61 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mongodb.kafka.connect.util.jmx.internal;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+
+public class AdditiveMetric implements Metric {
```
rozza, line 21: I don't full grok how this works, can we add some docs to explain how it works and where / when it should be used.

rozza (commented):
I think this is looking great - this is just a quick pass, looking at through the code. I'm going to debug / step through it before adding any more feedback as somethings I haven't fully grokked yet, which will come apparent when stepping through the code.

One question I have is do we have any idea how expensive adding this JMX reporting is?

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
```diff
@@ -85,24 +86,44 @@ void stop() {
...
       for (List<MongoProcessedSinkRecordData> batch : batches) {
         bulkWriteBatch(batch);
       }
     }
-    statistics.putTaskTime(taskTime);
-    lastTaskInvocation = statistics.startTimer();
+    statistics.getTaskInvocations().sample(taskTime.getElapsedTime(TimeUnit.MILLISECONDS));
+    lastTaskInvocation = Timer.start();
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug(statistics.getName() + ": " + statistics.toJSON());
```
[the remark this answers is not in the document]
katcharov: I wrote it this way because `toJSON` itself seemed relatively expensive, compared to the concatenation we'd be trying to save on here.

In response to your question about performance, I think it's negligible. We never increment counters in a tight loop. It's effectively an `AtomicLong` increment per stat. Usually this is once per poll, though iirc because of batching, retries, and the like, it might go up to "a few times" per poll.

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/AdditiveMetric.java
katcharov: Docs added. Perhaps the name `AdditiveMetric` isn't very clear, but I'm not sure what else to call this. The name metric itself might also be confusing - there are multiple values (MBean attributes) that arise out of a single measurement (`sample(long)`).

src/main/java/com/mongodb/kafka/connect/sink/StartedMongoSinkTask.java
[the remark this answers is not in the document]
katcharov: Added a small comment.

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
```diff
@@ -0,0 +1,133 @@
...
+          return mBeanNameToTry;
+        } catch (InstanceAlreadyExistsException e) {
+          LOGGER.warn("MBean name conflict " + mBeanNameToTry, e);
+          mBeanNameToTry = mBeanName + "-v" + NEXT_ID.getAndAdd(1);
+        }
+      }
+    } catch (Exception e) {
+      // JMX might not be available
+      LOGGER.warn("Unable to register MBean " + mBeanName, e);
+      return null;
```
rozza, line 75: Is returning `null` handled / valid here? If I registerMBean and it returns null - will unregistering it cause a NPE instead of being a Noop?

rozza (changes requested):
Couple of small nits.

Once done LGTM

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
jyemin: unregisterMBean seems to handle unregistering a not-registered MBean.  It both checks for registration (line 83) and swallows exceptions (line 86)

jyemin (approved):
LGTM

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
rozza: Ack - resolving.

rozza (approved):

src/main/java/com/mongodb/kafka/connect/util/jmx/internal/MBeanServerUtils.java
katcharov: The logger uses the mbean name, so this would log null in unusual circumstances. I returned an adjusted name.

PR #115, opened 2022-08-04
Log the reason why the heartbeat manager does not produce heartbeats

katcharov:
KAFKA-318

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/heartbeat/HeartbeatManager.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/heartbeat/HeartbeatManager.java
@@ -61,36 +61,43 @@ public class HeartbeatManager {
 
   public Optional<SourceRecord> heartbeat() {
     if (cursor == null) {
+      LOGGER.debug("Returning no heartbeat: null cursor");
       return Optional.empty();
     }
-
     long currentMS = time.milliseconds();
     long timeSinceHeartbeatMS = currentMS - lastHeartbeatMS;
-
-    if (canCreateHeartbeat && timeSinceHeartbeatMS > heartbeatIntervalMS) {
-      lastHeartbeatMS = currentMS;
-      return Optional.ofNullable(cursor.getResumeToken())
-          .map(
-              r -> {
-                String resumeToken = r.toJson();
-                if (!resumeToken.equals(lastResumeToken)) {
-                  LOGGER.info("Generating heartbeat event. {}", resumeToken);
-                  Map<String, String> sourceOffset = new HashMap<>();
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/kafka/connect/source/heartbeat/HeartbeatManager.java
```diff
@@ -61,36 +61,43 @@ public HeartbeatManager(
...
-                      heartbeatTopicName,
-                      Schema.STRING_SCHEMA,
-                      resumeToken,
-                      Schema.OPTIONAL_BYTES_SCHEMA,
-                      null);
-                }
-                return null;
-              });
+    if (!canCreateHeartbeat) {
+      LOGGER.debug("Returning no heartbeat: canCreateHeartbeat is false");
```
jyemin, line 70: Can we instead log the reason that `canCreateHeartbeat` is false?  It's not clear what the semantics of this variable is based just on its name.

src/main/java/com/mongodb/kafka/connect/source/heartbeat/HeartbeatManager.java
```diff
@@ -61,36 +61,43 @@ public HeartbeatManager(
...
+      LOGGER.debug("Returning no heartbeat: cursor resumeToken is null");
+      return Optional.empty();
+    }
+    String resumeToken = resumeTokenBson.toJson();
+    if (resumeToken.equals(lastResumeToken)) {
+      LOGGER.debug("Returning no heartbeat: same resumeToken");
+      return Optional.empty();
     }
-    return Optional.empty();
+    LOGGER.info("Generating heartbeat event. {}", resumeToken);
```
jyemin, line 88: What's the rationale for info level here?

jyemin (changes requested):

src/main/java/com/mongodb/kafka/connect/source/heartbeat/HeartbeatManager.java
katcharov: The message is preserved from the original happy-path code at line 77. I left it alone, though I can see how it might be argued to be either info or debug.

src/main/java/com/mongodb/kafka/connect/source/heartbeat/HeartbeatManager.java
katcharov: This couples the message to the canCreateHeartbeat logic, which is in a different method (the constructor). But it seems like it should be in this method, since this is where the rest of the checks are. I moved it.

jyemin (approved):

PR #117, opened 2022-09-09
SLF4J usage issues

jchipmunk:
Premature `toString()` invocation and string concatenation may cause performance issues in high-load applications.

[KAFKA-328](https://jira.mongodb.org/browse/KAFKA-328)


katcharov:
Good catch! Thank you for the PR.

PR #121, opened 2022-10-20
Add the new `change.stream.full.document.before.change` config property

stIncMale:
This new property is equivalent to [`ChangeStreamIterable.fullDocumentBeforeChange`](https://mongodb.github.io/mongo-java-driver/4.7/apidocs/mongodb-driver-sync/com/mongodb/client/ChangeStreamIterable.html#fullDocumentBeforeChange(com.mongodb.client.model.changestream.FullDocumentBeforeChange)) in the driver API.

**Problems**:

- ~~4 tests in `MongoSourceTaskTest` accidentally became integration tests (they need a running MongoDB). These tests need to be refactored.~~
  - **Solution**: rename `MongoSourceTaskTest` to `MongoSourceTaskIntegrationJunkTest` and move it to integration tests. These tests are irredeemable and need to be rewritten to become unit tests.

[KAFKA-308](https://jira.mongodb.org/browse/KAFKA-308)

### DIFF
--- a/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationJunkTest.java
+++ b/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationJunkTest.java
@@ -22,8 +22,6 @@ import static com.mongodb.kafka.connect.source.MongoSourceConfig.CONNECTION_URI_
 import static com.mongodb.kafka.connect.source.MongoSourceConfig.DATABASE_CONFIG;
 import static com.mongodb.kafka.connect.source.MongoSourceConfig.FULL_DOCUMENT_CONFIG;
 import static com.mongodb.kafka.connect.source.MongoSourceConfig.PIPELINE_CONFIG;
-import static com.mongodb.kafka.connect.source.SourceTestHelper.TEST_COLLECTION;
-import static com.mongodb.kafka.connect.source.SourceTestHelper.TEST_DATABASE;
 import static com.mongodb.kafka.connect.util.jmx.internal.MBeanServerUtils.getMBeanAttributes;
 import static java.lang.String.format;
 import static java.util.Collections.singletonMap;
@@ -33,6 +31,7 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -64,8 +63,15 @@ import com.mongodb.client.model.changestream.FullDocument;
 import com.mongodb.event.CommandFailedEvent;
 import com.mongodb.event.CommandSucceededEvent;
 
+import com.mongodb.kafka.connect.util.jmx.SourceTaskStatistics;
+
+/**
+ * This class contains tests that are supposed to be unit tests, but because of how these tests were
+ * written originally, they became integration tests as a result of refactoring {@link
+ * MongoSourceConfig}, and need to be completely rewritten to remain unit tests.
+ */
 @ExtendWith(MockitoExtension.class)
-class MongoSourceTaskTest {
+class MongoSourceTaskIntegrationJunkTest {
 
   @Mock private MongoClient mongoClient;
   @Mock private MongoDatabase mongoDatabase;
@@ -76,6 +82,8 @@ class MongoSourceTaskTest {
   @Mock private SourceTaskContext context;
   @Mock private OffsetStorageReader offsetStorageReader;
 
+  private static final String TEST_DATABASE = "myDB";
+  private static final String TEST_COLLECTION = "myColl";
   private static final BsonDocument RESUME_TOKEN = BsonDocument.parse("{resume: 'token'}");
   private static final Map<String, Object> OFFSET = singletonMap("_id", RESUME_TOKEN.toJson());
 
@@ -88,6 +96,7 @@ class MongoSourceTaskTest {
     cfgMap.put(DATABASE_CONFIG, TEST_DATABASE);
     cfgMap.put(COLLECTION_CONFIG, TEST_COLLECTION);
     MongoSourceConfig cfg = new MongoSourceConfig(cfgMap);
+    task.start(cfgMap);
 
     when(mongoClient.getDatabase(TEST_DATABASE)).thenReturn(mongoDatabase);
     when(mongoDatabase.getCollection(TEST_COLLECTION)).thenReturn(mongoCollection);
@@ -95,7 +104,7 @@ class MongoSourceTaskTest {
     when(changeStreamIterable.withDocumentClass(RawBsonDocument.class)).thenReturn(mongoIterable);
     when(mongoIterable.cursor()).thenReturn(mongoCursor);
 
-    task.createCursor(cfg, mongoClient);
+    task.startedTask().createCursor(cfg, mongoClient);
 
     verify(mongoClient, times(1)).getDatabase(TEST_DATABASE);
     verify(mongoDatabase, times(1)).getCollection(TEST_COLLECTION);
@@ -114,7 +123,7 @@ class MongoSourceTaskTest {
     when(changeStreamIterable.withDocumentClass(RawBsonDocument.class)).thenReturn(mongoIterable);
     when(mongoIterable.cursor()).thenReturn(mongoCursor);
 
-    task.createCursor(cfg, mongoClient);
+    task.startedTask().createCursor(cfg, mongoClient);
 
     verify(mongoClient, times(1)).getDatabase(TEST_DATABASE);
     verify(mongoDatabase, times(1)).getCollection(TEST_COLLECTION);
@@ -146,7 +155,7 @@ class MongoSourceTaskTest {
 
     task.initialize(context);
     when(context.offsetStorageReader()).thenReturn(offsetStorageReader);
-    when(offsetStorageReader.offset(task.createPartitionMap(cfg))).thenReturn(OFFSET);
+    when(offsetStorageReader.offset(MongoSourceTask.createPartitionMap(cfg))).thenReturn(OFFSET);
 
     when(mongoClient.getDatabase(TEST_DATABASE)).thenReturn(mongoDatabase);
     when(mongoDatabase.getCollection(TEST_COLLECTION)).thenReturn(mongoCollection);
@@ -158,7 +167,7 @@ class MongoSourceTaskTest {
     when(changeStreamIterable.withDocumentClass(RawBsonDocument.class)).thenReturn(mongoIterable);
     when(mongoIterable.cursor()).thenReturn(mongoCursor);
 
-    task.createCursor(cfg, mongoClient);
+    task.startedTask().createCursor(cfg, mongoClient);
 
     verify(mongoClient, times(1)).getDatabase(TEST_DATABASE);
     verify(mongoDatabase, times(1)).getCollection(TEST_COLLECTION);
@@ -169,6 +178,8 @@ class MongoSourceTaskTest {
     verify(changeStreamIterable, times(1)).startAfter(RESUME_TOKEN);
     verify(changeStreamIterable, times(1)).withDocumentClass(RawBsonDocument.class);
     verify(mongoIterable, times(1)).cursor();
+
+    task.stop();
   }
 
   @Test
@@ -179,13 +190,14 @@ class MongoSourceTaskTest {
     cfgMap.put(CONNECTION_URI_CONFIG, "mongodb://localhost");
     cfgMap.put(DATABASE_CONFIG, TEST_DATABASE);
     MongoSourceConfig cfg = new MongoSourceConfig(cfgMap);
+    task.start(cfgMap);
 
     when(mongoClient.getDatabase(TEST_DATABASE)).thenReturn(mongoDatabase);
     when(mongoDatabase.watch()).thenReturn(changeStreamIterable);
     when(changeStreamIterable.withDocumentClass(RawBsonDocument.class)).thenReturn(mongoIterable);
     when(mongoIterable.cursor()).thenReturn(mongoCursor);
 
-    task.createCursor(cfg, mongoClient);
+    task.startedTask().createCursor(cfg, mongoClient);
 
     verify(mongoClient, times(1)).getDatabase(TEST_DATABASE);
     verify(mongoDatabase, times(1)).watch();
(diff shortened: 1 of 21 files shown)


src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
```diff
@@ -197,439 +107,104 @@ public String version() {
...
-                valueDocument.toJson()));
-      }
-      throw new DataException(errorMessage.get(), e);
-    }
+    return startedTask.poll();
   }
 
   @Override
-  @SuppressWarnings("try")
-  public synchronized void stop() {
```
stIncMale, line 447: Removed `synchronized` as `stop` cannot be called concurrently with itself, only with `poll`/`commit`, but `poll`/`commit` were neither `synchronized` nor used `synchronized (this)` blocks.

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
```diff
@@ -0,0 +1,648 @@
...
+  private static final String RESUME_POINT = "resume point";
+  private static final String NOT_FOUND = "not found";
+  private static final String DOES_NOT_EXIST = "does not exist";
+  private static final String INVALID_RESUME_TOKEN = "invalid resume token";
+  private static final String NO_LONGER_IN_THE_OPLOG = "no longer be in the oplog";
+
+  private final Supplier<SourceTaskContext> sourceTaskContextAccessor;
+  private final Time time;
+  private volatile boolean isRunning;
+  private boolean isCopying;
```
stIncMale, line 113: `isCopying` is now a plain field because there is no point in setting it to `false` in `stop`. And without writes from `stop`, this field is always accessed sequentially.

src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java
```diff
@@ -280,7 +280,7 @@ void testBsonSchemaAndValueProducer() {
         "Assert schema and value matches",
         () -> assertEquals(Schema.BYTES_SCHEMA.schema(), actual.schema()),
         // Ensure the data length is truncated.
-        () -> assertEquals(1071, ((byte[]) actual.value()).length),
+        () -> assertEquals(1712, ((byte[]) actual.value()).length),
```
stIncMale, line 283: @rozza I have no idea where the previous `1071` value came from, or what this assertion is. I put `1712` as this is the new value we have in this test as a result of changes. Should we just delete the assertion as it is useless, given that a reader cannot understand what this is, and doing what I did appears to be the only way to make schema changes, unless the reader has access to arcane knowledge?

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
```diff
@@ -0,0 +1,649 @@
...
+  private MongoChangeStreamCursor<? extends BsonDocument> cursor;
+  private final StatisticsManager statisticsManager;
+  private final InnerOuterTimer inTaskPollInConnectFrameworkTimer;
+
+  StartedMongoSourceTask(
+      final Supplier<SourceTaskContext> sourceTaskContextAccessor,
+      final MongoSourceConfig sourceConfig,
+      final MongoClient mongoClient,
+      @Nullable final MongoCopyDataManager copyDataManager,
+      final StatisticsManager statisticsManager) {
```
stIncMale, line 135: `StartedMongoSourceTask` was extracted from `MongoSourceTask`, it gets all its dependencies via a constructor, which makes it unit-testable.

src/test/java/com/mongodb/kafka/connect/source/StartedMongoSourceTaskTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.mongodb.client.ChangeStreamIterable;
+import com.mongodb.client.MongoChangeStreamCursor;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.model.changestream.FullDocumentBeforeChange;
+
+import com.mongodb.kafka.connect.source.statistics.JmxStatisticsManager;
+
+final class StartedMongoSourceTaskTest {
+  static final class FullDocumentBeforeChangeTest {
```
stIncMale, line 45: I am proposing the idea of grouping unit tests in different static nested classes by functionality, as this prevents cluttering the test state, set up and tear down methods. If you think this is something that should not be done, or that it may be done, but only when we have unit-tests for different parts of functionality, or something else, please express your thoughts.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -194,6 +196,17 @@ public class MongoSourceConfig extends AbstractConfig {
...
 
+  public static final String FULL_DOCUMENT_BEFORE_CHANGE_CONFIG =
+      "change.stream.full.document.before.change";
+  private static final String FULL_DOCUMENT_BEFORE_CHANGE_DISPLAY =
+      "The fullDocumentBeforeChange configuration.";
+  private static final String FULL_DOCUMENT_BEFORE_CHANGE_DOC =
+      "Specifies the pre-image configuration when creating a Change Stream.\n"
+          + "The pre-image is not available in source records published while copying existing data as a result of"
+          + " enabling `copy.existing`, and the pre-image configuration has no effect on copying.\n"
+          + "See https://www.mongodb.com/docs/manual/reference/method/db.collection.watch/ for more details and possible values.";
```
stIncMale, line 207: I linked the server docs here instead of listing and explaining all possible values, because, as always, duplicating the docs eventually leads to the duplicated docs becoming inconsistent and/or obsolete, which results in misleading users, which in turn, makes users to distrust the docs.

In the past, I gave examples of problems our approach introduced in the driver API (where the decision has been and stays "the convenience of reading docs in place is more important than having consistent docs users may trust"), and we have similar situations in Kafka. For example, the documentation for `change.stream.full.document` config property states that possible values are only `""`, `"default"`, `"updateLookup"`. The new values `"whenAvailable"` and `"required"` are omitted, as they were added later - what should users thing about this? Does the connector not support these values? Does it support them, but the authors forgot to specify them? If users actually trust the connector docs (they should not), they will not learn about these two new values, or any values that will be supported by the server in the future.

src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java
stIncMale: It is expected that the serialized value is longer, as it has more data now. But I know neither where the asserted value is supposed to come from, nor why the serialized value is expected to be truncated.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationJunkTest.java
```diff
@@ -64,8 +63,15 @@
...
+import com.mongodb.kafka.connect.util.jmx.SourceTaskStatistics;
+
+/**
+ * This class contains tests that are supposed to be unit tests, but because of how these tests were
+ * written originally, they became integration tests as a result of refactoring {@link
+ * MongoSourceConfig}, and need to be completely rewritten to remain unit tests.
+ */
 @ExtendWith(MockitoExtension.class)
-class MongoSourceTaskTest {
+class MongoSourceTaskIntegrationJunkTest {
```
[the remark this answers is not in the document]
stIncMale: On one hand we already have the `MongoSourceTaskIntegrationTest` class, and I did not want to move these tests as they don't belong there. On the other hand I wanted to give the class a name that suggests we should refactor the tests (btw, I have just mentioned this in KAFKA-234). I am open to use a different name, if you have one in mind.

The problems with these tests (even as they were before this PR) are:

- They appear to test multiple things within a single test, which makes some of them multi-page tests (see `testCreatesExpectedCollectionCursor`).
- Probably as a consequence of the previous problem, the tests have to call a magic `resetMocks` method instead of re-creating mocks in `setUp`/`tearDown` or within each test.
- The tests heavily rely on internal methods of the object under test. While it may be because the object under test has too many unrelated behaviors and needs to be divided in multiple objects, I expect that the more likely reason this happened is that `MongoSourceTask` is not unit-testable. While `MongoSourceTask` continues to be not unit-testable, the `StartedMongoSourceTask` that was extracted from it is. These tests need to be refactored to unit-test `StartedMongoSourceTask`.

src/main/java/com/mongodb/kafka/connect/source/statistics/JmxStatisticsManager.java
```diff
@@ -0,0 +1,69 @@
...
+package com.mongodb.kafka.connect.source.statistics;
+
+import com.mongodb.annotations.ThreadSafe;
+
+import com.mongodb.kafka.connect.util.jmx.SourceTaskStatistics;
+import com.mongodb.kafka.connect.util.jmx.internal.CombinedMongoMBean;
+import com.mongodb.kafka.connect.util.jmx.internal.MBeanServerUtils;
+
+@ThreadSafe
+public final class JmxStatisticsManager implements StatisticsManager {
```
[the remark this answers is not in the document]
stIncMale: What do you think should be done with the sink stats once they are in this class, and what will be achieved as a result?

src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java
katcharov: We should not delete the assertion, because it guards against unexpected changes to the value, even though we do not know exactly how to compute the value.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationJunkTest.java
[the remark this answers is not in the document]
katcharov: I do not think we should track the status via the name of the test. `MongoSourceTaskIntegrationTestTODONeedsRefactoring` would be better, but more obviously not the right thing. If the problems with the test are not obvious to a future maintainer, we might describe them in a comment, or a Jira ticket if we think we should spend time on this (but, if not now, then when?).

src/main/java/com/mongodb/kafka/connect/source/statistics/JmxStatisticsManager.java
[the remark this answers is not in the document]
katcharov: All the stats will be in one place, which is presumably the purpose of the JmxStatisticsManager.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationJunkTest.java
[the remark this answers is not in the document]
stIncMale: > we might describe them in a comment, or a Jira ticket if we think we should spend time on this

KAFKA-234 is such a ticket.

> if not now, then when?

When we decide to do KAFKA-234.

> MongoSourceTaskIntegrationTestTODONeedsRefactoring would be better, but more obviously not the right thing

We have to have a name different from `MongoSourceTaskIntegrationTest` because we already have that class. So despite this being not the right thing, do you want the class to be renamed as `MongoSourceTaskIntegrationTestTODONeedsRefactoring`?

src/main/java/com/mongodb/kafka/connect/source/statistics/JmxStatisticsManager.java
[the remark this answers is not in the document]
stIncMale: We discussed this in Zoom and agreed that there is nothing to do here.

katcharov (approved):

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationJunkTest.java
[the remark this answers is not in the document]
stIncMale: Done.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
rozza: Ack

src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java
rozza: Ack

src/test/java/com/mongodb/kafka/connect/source/StartedMongoSourceTaskTest.java
rozza: Anything that validates the connector and makes the code easier to read / come back to / consume sounds great to me. Do you want to add a separate ticket for this?

src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java
rozza: iirc it was added to ensure that the size is as expected - it really is testing documentToByteArray and its handling of RawBsonDocuments. So I agree it can be removed from here and is better suited to its own test and / or is already tested in BsonValueToSchemaAndValueTest

src/test/java/com/mongodb/kafka/connect/source/StartedMongoSourceTaskTest.java
stIncMale: I don't think we need to go and update existing tests, but rather I was curious whether it is fine to start using the approach when writing new tests (and the answer appears to be "yes"), if that seems helpful. As for the existing tests, we may regroup them when we touch them if, again, that seems beneficial in a specific case.

src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java
stIncMale: Removed the assertions.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest2.java
```diff
@@ -64,8 +63,15 @@
...
+import com.mongodb.kafka.connect.util.jmx.SourceTaskStatistics;
+
+/**
+ * This class contains tests that are supposed to be unit tests, but because of how these tests were
+ * written originally, they became integration tests as a result of refactoring {@link
+ * MongoSourceConfig}, and need to be completely rewritten to remain unit tests.
+ */
 @ExtendWith(MockitoExtension.class)
-class MongoSourceTaskTest {
+class MongoSourceTaskIntegrationTest2 {
```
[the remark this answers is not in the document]
stIncMale: What does it mean for a test to be verification test? It appears to me that all tests verify something. I'd prefer to leave the `...IntegrationTest2` name. It won't survive past KAFKA-234.

rozza (approved):
LGTM

PR #124, opened 2022-11-04
Use type Attribute for attributes in JMX mbeans

viruzzo:
The classes **`MongoMBean`** and **`CombinedMongoMBean`** in `com.mongodb.kafka.connect..util.jmx.internal` currently return metric values directly from their _`getAttribute()`_ method and inside the `AttributeList` from their _`getAttributes()`_ method, but the documentation for `AttributeList` specifically says (emphasis mine):

> For compatibility reasons, it is possible, **though highly discouraged**, to add objects to an AttributeList that are not instances of Attribute.

The current behavior breaks at the very least https://github.com/prometheus/jmx_exporter, since they ignore non-`Attribute` values in their metric export.

The proposed change makes it so that the `AttributeList` returned by _`getAttributes()`_ is filled with `Attribute`s wrapping the the metric values, and _`getAttribute()`_ returns an `Attribute` for coherence and general compliance with the interface's intentions.

Edit: bungled the integration test change, it's now fixed, apologies.

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/util/jmx/internal/CombinedMongoMBean.java
+++ b/src/main/java/com/mongodb/kafka/connect/util/jmx/internal/CombinedMongoMBean.java
@@ -48,7 +48,7 @@ public class CombinedMongoMBean implements DynamicMBean {
   public Object getAttribute(final String attribute)
       throws AttributeNotFoundException, MBeanException, ReflectionException {
     if (metricsMap.containsKey(attribute)) {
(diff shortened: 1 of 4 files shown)


viruzzo:
I can't see the evergreen result details above, but `./gradlew check ...` had passed in local (with a replica set) with the latest commit.

katcharov (approved):
Looks good, thank you for catching this.

There wasn't any issue with the code, evergreen was just acting up.

PR #125, opened 2022-11-05
Add the new `startup.mode.timestamp.start.at.operation.time` config property

stIncMale:
Instead of just introducing a new config property, I observed that the behavior introduced by it is mutually exclusive with the `copy.existing.*` behavior, and both define the connector behavior when there is no offset data. This observation should be reflected in the structure of our source configuration, if we want the configuration to be clear to a user (and us, the maintainers of the connector). So I did the following:

- ~~Introduced the `start` configuration property with two values `ignore_existing` and `copy_existing`. The latter is equivalent to setting `copy.existing = true`.~~
- ~~Deprecated `copy.existing`, which required refactoring the code we use for soft validation of configs. The property is still supported.~~
- ~~Refactored the use of the rest of the `copy.existing.*` properties such that it's clear from the code that they have effect only if `start = copy_existing`.~~
- ~~Introduced and implemented the new `ignore.existing.before.operation.time` property, which has effect only if `start = ignore_existing`.~~

- Introduced the `start.mode` configuration property with three values `latest`, `timestamp`, `copy_existing`. The former is equivalent to the current default behavior, the latter is equivalent to setting `copy.existing = true`.
- The `timestamp` value actuates the new `startup.mode.timestamp.start.at.operation.time` property.
- Deprecated `copy.existing*` properties, which required refactoring the code we use for soft validation of configs. The deprecated properties are still supported.
- Introduced `startup.mode.copy.existing.*` properties as replacements for the deprecated `copy.existing.*` properties.
- Refactored the use of the rest of the `*copy.existing*` properties such that it's clear from the code that they have effect only if `start.mode = copy_existing`.

KAFKA-330


src/main/java/com/mongodb/kafka/connect/sink/SinkConfigSoftValidator.java
```diff
@@ -32,58 +32,43 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.mongodb.annotations.Immutable;
-import com.mongodb.lang.Nullable;
-
-import com.mongodb.kafka.connect.util.VisibleForTesting;
-
-/** Logs information about potential configuration issues. */
-public final class SinkConfigSoftValidator {
```
stIncMale, line 68: This class was split into `ConfigSoftValidator` (contains the validation functionality) and `SinkConfigSoftValidator`/`SourceConfigSoftValidator` that contain specific business rules to check (unused, deprecated, conflicting properties).

src/main/java/com/mongodb/kafka/connect/util/config/ConfigSoftValidator.java
```diff
@@ -0,0 +1,329 @@
...
+    } else {
+      return "";
+    }
+  }
+
+  private ConfigSoftValidator() {}
+
+  /** A description of a set of obsolete configuration properties. */
+  @Immutable
+  public static final class ObsoletePropertiesSet {
```
stIncMale, line 175: The `ObsoletePropertiesSet` class is a new piece of code in `ConfigSoftValidator` (the content of `ConfigSoftValidator` was originally in `SinkConfigSoftValidator`).

src/main/java/com/mongodb/kafka/connect/util/config/ConfigSoftValidator.java
```diff
@@ -0,0 +1,329 @@
...
+
+/** Logs information about potential configuration issues. */
+public final class ConfigSoftValidator {
+  /**
+   * This method is used in our implementation of the {@link ConfigDef#validateAll(Map)} method
+   * because the original implementation silently disregards any undefined configuration properties.
+   *
+   * @param propertyNames See the parameter description in {@link #strippedPropertyName(String)}.
+   */
+  public static void logObsoleteProperties(
```
stIncMale, line 53: The implementation of the `logObsoleteProperties` has changed from what we originally had in `SinkConfigSoftValidator` before its code was moved here. The `ObsoletePropertiesSet` class is also a new piece of code. The rest of the `ConfigSoftValidator` was just moved here from `SinkConfigSoftValidator`.

rozza:
I think: `start.at.operation.time` would be easier to understand / discover and document than a totally new config `ignore.existing.before.operation.time`. But I get the logic for adding it, so will defer to Robert on config names / deprecating copy.existing.

For the config is there need for the increment or will just seconds since epoch suffice?  Much easier to configure for users if that is the case. Not 100% sure of the use case.  Setting `startAtOperationTime` being to `new BsonTimeStamp(0)` will start at the beginning of the oplog - rather than from the end. Is this something we should default to? 

@RWaltersMA any thoughts?

Jiabao-Sun:
For the config names, how about use `startup.mode` which provides three mode for user to choose.
- full:  `copy.existing = true`
- earliest:  `copy.existing = false` and from the start at the beginning of the oplog
- latest: `copy.existing = false`
- timestamp:  `copy.existing = false` and from the specific timestamp. And an extra option`startup.at.operation.time` is needed for this startup mode.

Jiabao-Sun (commented):
https://github.com/mongodb/mongo-kafka/blob/2662926a35411844bd9bcd474ea71a91efcde3d2/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L466
The log message `copy.existing=true` may be modified as well.

https://github.com/mongodb/mongo-kafka/blob/2662926a35411844bd9bcd474ea71a91efcde3d2/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L440-L457
The `startAtOperationTime` was supported after MongoDB 4.0, we may check this error here rather than return a null ChangeStreamCursor that will dive into an infinite loop.

```
com.mongodb.MongoCommandException: Command failed with error 40415 (Location40415):
 'BSON field '$changeStream.startAtOperationTime' is an unknown field.' on server 
localhost:27017. The full response is {"operationTime": {"$timestamp": {"t": 1667838442, "i": 
1}}, "ok": 0.0, "errmsg": "BSON field '$changeStream.startAtOperationTime' is an unknown 
field.", "code": 40415, "codeName": "Location40415", "$clusterTime": {"clusterTime": 
{"$timestamp": {"t": 1667838442, "i": 1}}, "signature": {"hash": {"$binary": {"base64": 
"AAAAAAAAAAAAAAAAAAAAAAAAAAA=", "subType": "00"}}, "keyId": 0}}}
```

stIncMale:
Thanks, @Jiabao-Sun, [those](https://github.com/mongodb/mongo-kafka/pull/125#pullrequestreview-1170772575) are very helpful catches. Done.

As for your [proposal](https://github.com/mongodb/mongo-kafka/pull/125#issuecomment-1305796075), something like that was my initial approach (though I did not consider exposing something equivalent to `earliest` explicitly as it can be achieved via what is called `timestamp` in your proposal). However, I did not like that `timestamp` has to either require a user to specify `timestamp.start.at.operation.time`<sup>1</sup> (I'd like to avoid such a behavior when possible), or has to explain that it is equivalent to `startup.mode = latest` if `timestamp.start.at.operation.time` is not specified. I can live with that if that is fine by the rest of the team because this gives us the desired `timestamp.start.at.operation.time` property name.

---

<sup>1</sup> By prefixing dependent property names with the `start`/`startup.mode` value they are actuated by, we express the relation via naming, which is why `timestamp.start.at.operation.time` is preferrable to `startup.at.operation.time` (for the same reason the value `full` should rather be `copy_existing`). It would have been even better to additionally include the `start`/`startup.mode` property name in the prefix, but this approach cannot be reconciled with all the `copy.existing.*` properties without deprecating them all, which I anticipate, would have made getting the rest of the team onboard with the approach even harder.

rozza:
Just to note: Kafka has the following option: `auto.offset.reset` which can be `earliest` or `latest` which is the equivalent to setting setting `startAtOperationTime(new BsonTimeStamp(0))` and not setting a `startAtOperationTime` value.

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.kafka.common.config.ConfigException;
+
+import org.bson.BsonTimestamp;
+import org.bson.json.JsonReader;
+
+public final class BsonTimestampParser {
+  public static final String FORMAT_DESCRIPTION =
```
rozza, line 28: Like this :)

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
```diff
@@ -0,0 +1,59 @@
...
+  public static BsonTimestamp parse(final String propertyName, final String propertyValue)
+      throws ConfigException {
+    List<RuntimeException> exceptions = new ArrayList<>();
+    try {
+      return new BsonTimestamp(Integer.parseInt(propertyValue), 0);
+    } catch (RuntimeException e) {
+      exceptions.add(e);
+    }
+    try {
+      return new BsonTimestamp(Math.toIntExact(Instant.parse(propertyValue).getEpochSecond()), 0);
```
rozza, line 43: Should we log a warning or error if they provide milliseconds eg: `2018-11-30T18:35:24.12Z`?

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
```diff
@@ -0,0 +1,59 @@
...
+import org.bson.json.JsonReader;
+
+public final class BsonTimestampParser {
+  public static final String FORMAT_DESCRIPTION =
+      "Must be either an integer number of seconds since the Epoch in the decimal format (example: 30),"
+          + " or an instant in the ISO-8601 format (example: '1970-01-01T00:00:30Z'),"
+          + " or a BSON Timestamp in the canonical extended JSON (v2) format"
+          + " (example: '{\"$timestamp\": {\"t\": 30, \"i\": 0}}').";
+
+  public static BsonTimestamp parse(final String propertyName, final String propertyValue)
```
rozza, line 34: Lets add a unit test for this.

rozza (changes requested):

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
RWaltersMA: I assume ms are not supported ?

rozza:
We can always add documentation in the main docs for how to achieve earliest / latest semantics.

A downside to deprecation is it just generates a log message, so is not very obvious to end users, especially as Kafka connect can generate a lot of log messages.  However, the approach taken still supports these deprecated configs then I'm happy if we can find a cleaner way to express configuration names.

Note suffixing dependant property names should be preferred as; it naturally groups values by the naming convention and is the established convention in Kafka - eg: SMT configuration: https://docs.confluent.io/platform/current/connect/transforms/flatten.html#flatten

src/main/java/com/mongodb/kafka/connect/sink/SinkConfigSoftValidator.java
rozza: Ack

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -250,56 +267,161 @@ public class MongoSourceConfig extends AbstractConfig {
           + "watched.";
   private static final String COLLECTION_DEFAULT = EMPTY_STRING;
 
-  public static final String COPY_EXISTING_CONFIG = "copy.existing";
+  public static final String STARTUP_MODE_CONFIG = "startup.mode";
+  private static final String STARTUP_MODE_CONFIG_DISPLAY =
+      "The behavior when there is no offset data.";
+  private static final String STARTUP_MODE_CONFIG_DOC =
+      format(
+          "A change stream cannot be resumed if there is no offset data."
```
rozza, line 275: The first sentence is a little confusing in its relationship to the config.

Suggest simplifying the initial description to something like "The initial connector start up mode."

Then later put a note stating that once the connector has been started and change stream offset data is stored any restarts will continue from that offset.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -250,56 +267,161 @@ public class MongoSourceConfig extends AbstractConfig {
           + "watched.";
   private static final String COLLECTION_DEFAULT = EMPTY_STRING;
 
-  public static final String COPY_EXISTING_CONFIG = "copy.existing";
+  public static final String STARTUP_MODE_CONFIG = "startup.mode";
+  private static final String STARTUP_MODE_CONFIG_DISPLAY =
+      "The behavior when there is no offset data.";
```
rozza, line 272: Suggest: The initial start up behavior when there is no offset data.

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
```diff
@@ -85,10 +89,10 @@
 import com.mongodb.kafka.connect.util.time.InnerOuterTimer.InnerTimer;
 
 final class StartedMongoSourceTask implements AutoCloseable {
-  private static final Logger LOGGER = LoggerFactory.getLogger(MongoSourceTask.class);
   private static final String FULL_DOCUMENT = "fullDocument";
   private static final int NAMESPACE_NOT_FOUND_ERROR = 26;
   private static final int ILLEGAL_OPERATION_ERROR = 20;
+  private static final int UNKNOWN_FIELD_ERROR = 40415;
```
rozza, line 95: Is this the error for setting startAtOperationTime pre 4.0?

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -250,56 +267,161 @@ public class MongoSourceConfig extends AbstractConfig {
...
+  private static final String STARTUP_MODE_CONFIG_DOC =
+      format(
+          "A change stream cannot be resumed if there is no offset data."
+              + " In this case the connector may either ignore all/some existing source data,"
+              + " or may at first copy all existing source data and then continue with processing new data."
+              + " Possible values are %s."
+              + "\n- 'latest' is the default value. The connector ignores all existing source data and processes only new data."
+              + "\n- 'timestamp' actuates 'startup.mode.timestamp.*' properties."
```
rozza, line 280: Add note that timestamp requires MongoDB 4.0 or greater.

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
```diff
@@ -435,6 +455,10 @@ private MongoChangeStreamCursor<? extends BsonDocument> tryCreateCursor(
             e.getErrorCode(),
             e.getErrorMessage());
         throw new ConnectException("Illegal $changeStream operation", e);
+      } else if (e.getErrorCode() == UNKNOWN_FIELD_ERROR) {
+        String msg = format("Invalid operation: %s %s.", e.getErrorMessage(), e.getErrorCode());
```
rozza, line 459: This error message is likely to confuse users - this should only happen on MongoDB < 4.0 with a timestamp set. Can you add more information to help users determine the root cause of issue and how to fix?

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
```diff
@@ -0,0 +1,67 @@
...
+import com.mongodb.lang.Nullable;
+
+public final class BsonTimestampParser {
+  public static final String FORMAT_DESCRIPTION =
+      "Must be either an integer number of seconds since the Epoch in the decimal format (example: 30),"
+          + " or an instant in the ISO-8601 format with one second precision (example: '1970-01-01T00:00:30Z'),"
+          + " or a BSON Timestamp in the canonical extended JSON (v2) format"
+          + " (example: '{\"$timestamp\": {\"t\": 30, \"i\": 0}}').";
+
+  public static BsonTimestamp parse(
```
rozza, line 37: This class should have its own stand alone unit tests. Its not obvious that there are existing tests in `MongoSourceConfigTest` - they probably can just be moved over to `BsonTimestampParserTest`.

This way in the future maintainers will find it easier should any changes be required.

rozza (changes requested):
Looking good - couple of nits and suggestions.

katcharov (approved):
LGTM. I like the approach taken (of having mutually exclusive configs represented under a single property).

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
stIncMale: What does "initial" means here and [below](https://github.com/mongodb/mongo-kafka/pull/125#discussion_r1018830236)? The only meaning I can think of based on your suggestion is "when there is no offset data". If that's the case, then what's the purpose of introducing a new term "initial start up" given that we don't use it anywhere else, including the documentation in question?

> Then later put a note stating that once the connector has been started and change stream offset data is stored any restarts will continue from that offset.

Note that `offset.partition.name` explains nothing about what "offset" means. This seems reasonable given that it's the Connect Framework term, not a connector-specific term.

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
stIncMale: Yes. But it may also happen with other fields we may introduce in the future, i.e., we should not make the error message specific to `startAtOperationTime`.

stIncMale (commented):
For some reason GitHub interprets my comments as part of a review (which I have never started) and shows them as "pending", thus requiring me to add a review.

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
stIncMale: Done.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
stIncMale: I updated the doc, please take a look.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
stIncMale: Done.

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
stIncMale: Done.

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
stIncMale: This thread is a duplicate of https://github.com/mongodb/mongo-kafka/pull/125#discussion_r1016544358. Done.

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
rozza: One is an integration test, the other a unit test. Ideally, you'd only need the unit test for  testing multiple values and one integration test.

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
rozza: *note: Seconds are supported and there's a test case for that.

rozza (approved):
LGTM!

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
stIncMale: Oh, I wanted to say *nano*seconds :face_palm Thanks for noticing, Ross!

src/main/java/com/mongodb/kafka/connect/util/config/BsonTimestampParser.java
stIncMale: I agree. Removed some tests from `MongoSourceConfigTest`.

PR #128, opened 2023-01-03
Combine compatible schemas for array elements

jyemin:
1. Where two struct types differ only where one has a field that the other does not.  In this case, the extra field is added to the combined schema
 2. Where two struct types differ only where one has a field that the other has a field whose value is null in the source document. In this case, the extra field is added to the combined schema with the type of the field that has the non-null value.
 3. Where two array types differ only in that one of the arrays is empty. In this case, the value schema for the empty array is changed to the one for the non-empty array

KAFKA-343

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
@@ -16,26 +16,37 @@
 
 package com.mongodb.kafka.connect.source.schema;
 
+import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 
 import org.apache.kafka.connect.data.Decimal;
+import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Timestamp;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import org.bson.BsonArray;
 import org.bson.BsonDocument;
 import org.bson.BsonValue;
 
+import com.mongodb.kafka.connect.source.MongoSourceTask;
+
 public final class BsonDocumentToSchema {
 
+  static final Logger LOGGER = LoggerFactory.getLogger(MongoSourceTask.class);
+
+  private static final String SENTINEL_FOR_NULL_OR_EMPTY = "";
   private static final String ID_FIELD = "_id";
   private static final Schema DEFAULT_INFER_SCHEMA_TYPE = Schema.OPTIONAL_STRING_SCHEMA;
   public static final String DEFAULT_FIELD_NAME = "default";
 
   public static Schema inferDocumentSchema(final BsonDocument document) {
-    return createSchemaBuilder(DEFAULT_FIELD_NAME, document).required().build();
+    Schema schema = createSchemaBuilder(DEFAULT_FIELD_NAME, document).required().build();
+    return normalizeSchema(schema);
   }
 
   private static Schema inferDocumentSchema(final String fieldPath, final BsonDocument document) {
@@ -80,24 +91,39 @@ public final class BsonDocumentToSchema {
       case DOCUMENT:
         return inferDocumentSchema(fieldPath, bsonValue.asDocument());
       case ARRAY:
-        List<BsonValue> values = bsonValue.asArray().getValues();
-        Schema firstItemSchema =
-            values.isEmpty() ? DEFAULT_INFER_SCHEMA_TYPE : inferSchema(fieldPath, values.get(0));
-        if (values.isEmpty()
-            || values.stream()
-                .anyMatch(bv -> !Objects.equals(inferSchema(fieldPath, bv), firstItemSchema))) {
-          return SchemaBuilder.array(DEFAULT_INFER_SCHEMA_TYPE).name(fieldPath).optional().build();
+        BsonArray bsonArray = bsonValue.asArray();
+        if (bsonArray.isEmpty()) {
+          return SchemaBuilder.array(
+                  SchemaBuilder.string()
+                      .optional()
+                      // Sentinel value added to detect the empty array case.
+                      .defaultValue(SENTINEL_FOR_NULL_OR_EMPTY)
+                      .build())
+              .name(fieldPath)
+              .optional()
+              .build();
         }
-        return SchemaBuilder.array(inferSchema(fieldPath, bsonValue.asArray().getValues().get(0)))
-            .name(fieldPath)
-            .optional()
-            .build();
+        Schema combinedSchema = inferSchema(fieldPath, bsonArray.get(0));
+        for (int i = 1; i < bsonArray.size(); i++) {
+          combinedSchema = combine(combinedSchema, inferSchema(fieldPath, bsonArray.get(i)));
+          if (combinedSchema == null) {
+            break;
+          }
+        }
+        return combinedSchema == null
+            ? SchemaBuilder.array(DEFAULT_INFER_SCHEMA_TYPE).name(fieldPath).optional().build()
+            : SchemaBuilder.array(combinedSchema).name(fieldPath).optional().build();
       case BINARY:
         return Schema.OPTIONAL_BYTES_SCHEMA;
       case SYMBOL:
       case STRING:
-      case NULL:
         return Schema.OPTIONAL_STRING_SCHEMA;
+      case NULL:
+        return SchemaBuilder.string()
+            .optional()
+            // Sentinel value added to detect the empty array case.
+            .defaultValue(SENTINEL_FOR_NULL_OR_EMPTY)
+            .build();
       case OBJECT_ID:
       case REGULAR_EXPRESSION:
       case DB_POINTER:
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -80,24 +91,39 @@ private static Schema inferSchema(final String fieldPath, final BsonValue bsonVa
...
+                      .build())
+              .name(fieldPath)
+              .optional()
+              .build();
         }
-        return SchemaBuilder.array(inferSchema(fieldPath, bsonValue.asArray().getValues().get(0)))
-            .name(fieldPath)
-            .optional()
-            .build();
+        Schema combinedSchema = inferSchema(fieldPath, bsonArray.get(0));
```
rozza, line 106: Nice I think we can do this via a reduce - similar to how Spark does it.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -111,6 +137,134 @@ private static Schema inferSchema(final String fieldPath, final BsonValue bsonVa
...
+  private static boolean isSentinelValueSet(final Schema schema) {
+    return schema.type() == Schema.Type.STRING
+        && SENTINEL_FOR_NULL_OR_EMPTY.equals(schema.defaultValue());
+  }
+
+  /**
+   * This method normalizes the schema that had been denormalized due to the array element
+   * schema-combining logic applied as part of schema building process.
+   */
+  private static Schema normalizeSchema(final Schema schema) {
```
rozza, line 222: This shouldn't be needed as fields are already sorted (see createSchemaBuilder) so this looks to be duplicating work.

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
```diff
@@ -0,0 +1,213 @@
...
+                new BsonArray(
+                    asList(
+                        new BsonDocument("a", new BsonArray(asList(new BsonInt32(1)))),
+                        new BsonDocument("a", new BsonArray(asList(new BsonString("foo")))))));
+
+    Schema schema = inferDocumentSchema(document);
+    assertEquals(STRUCT, schema.type());
+
+    Schema outerArraySchema = schema.field("outerArray").schema();
+    assertEquals(STRING, outerArraySchema.valueSchema().type());
```
rozza, line 193: This looks off as only the field "a" array type is incompatible 

So I'd expect:

```
    Schema schema = inferDocumentSchema(document);
    assertEquals(STRUCT, schema.type());

    Schema outerArraySchema = schema.field("outerArray").schema();
    assertEquals(ARRAY, outerArraySchema.type());

    Schema outerArrayValueSchema = outerArraySchema.valueSchema();
    assertEquals(STRUCT, outerArrayValueSchema.type());

    Schema fieldSchema = outerArrayValueSchema.field("a").schema();
    assertEquals(ARRAY, fieldSchema.type());
    assertEquals(STRING, fieldSchema.valueSchema().type());
```

rozza (changes requested):
I found the tests differed from the existing tests in `BsonValueToSchemaAndValueTest` but I agree having unit tests that just handled the schema inference explicitly is worth it.

I had a look and reviewed how the Spark connector does similar (although its pattern is about combining schema for multiple documents, rather than for each document as per Kafka).

So I made some changes here in on my fork: https://github.com/rozza/mongo-kafka/commit/e4311ba135fa666e2ab642b17cfaaed017d18f11

If your happy I can merge into your branch.

jyemin:
@rozza had a look at your branch.  I like the use of the sentinel and of `Stream#reduce`.  But I don't think we can get rid of normalize entirely, as the combine method does not ensure alpha ordering of the fields.  I proved that by modifying one of the existing tests (see the next commit).  But we can reduce the scope of normalization by only normalizing the array schema, now that we don't have to worry about the sentinel value anymore.  I'll give that a try.

jyemin:
Yeah, that works.  Let me know what you think.

jyemin:
Reverted that since I don't have your sentinel change yet.  But it should work once combined with your work.

rozza:
Good catch updated my branch to include the test cases: 
https://github.com/rozza/mongo-kafka/compare/master...k343

Opted to fix ordering in the combine stage as we go along rather than a final step. 
See: https://github.com/rozza/mongo-kafka/commit/562498b503c869b2feffafb864f70f488f3eda8d

jyemin:
I like it!  Can you add to this PR?

rozza:
Done!

rozza (approved):
LGTM - but please also review :)

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -18,21 +18,26 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.bson.BsonDocument;
 import org.bson.BsonValue;
 
 public final class BsonDocumentToSchema {
-
+  public static final String DEFAULT_FIELD_NAME = "default";
+  private static final Logger LOGGER = LoggerFactory.getLogger(BsonDocumentToSchema.class);
```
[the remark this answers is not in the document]
rozza: That is the pattern that is used throughout the library. I believe as the namespaces are hierarchical users can configure loggers with namespace granularity.

jyemin:
Checkstyle is complaining about something.

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
jyemin: Is this still an issue?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -18,21 +18,26 @@
...
   private static final String ID_FIELD = "_id";
   private static final Schema DEFAULT_INFER_SCHEMA_TYPE = Schema.OPTIONAL_STRING_SCHEMA;
-  public static final String DEFAULT_FIELD_NAME = "default";
+  private static final Schema SENTINEL_STRING_TYPE =
+      SchemaBuilder.type(Schema.Type.STRING).optional().build();
```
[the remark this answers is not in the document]
rozza: Done

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -111,6 +111,90 @@ private static Schema inferSchema(final String fieldPath, final BsonValue bsonVa
     }
   }
 
+  private static Schema combine(final Schema firstSchema, final Schema secondSchema) {
+    if (isSentinelValueSet(firstSchema)) {
```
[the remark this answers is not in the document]
rozza: The initial value is always a sentinel because: 

`.reduce(SENTINEL_STRING_TYPE, BsonDocumentToSchema::combine)`

So we have to add a check for the initial value being the sentinel.

The rest falls through the logic.

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
```diff
@@ -0,0 +1,244 @@
...
+import org.apache.kafka.connect.data.Decimal;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.SchemaBuilder;
+import org.apache.kafka.connect.data.Timestamp;
+import org.junit.jupiter.api.Test;
+
+import org.bson.BsonDocument;
+
+@SuppressWarnings("ArraysAsListWithZeroOrOneArgument")
+public class BsonDocumentToSchemaTest {
```
jyemin, line 32: I ran these new tests with coverage enabled and most of the `BsonDocumentToSchema#combineFieldSchema` method is not covered by tests.  So I'm not clear why the tests that were added in the initial commit have been removed, as I think those did cover that method.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -111,6 +111,90 @@ private static Schema inferSchema(final String fieldPath, final BsonValue bsonVa
...
+    Field _id2 = secondSchema.field(ID_FIELD);
+    if (_id1 != null || _id2 != null) {
+      builder.field(ID_FIELD, combineFieldSchema(_id1, _id2));
+    }
+    // Combine other fields in name order
+    Stream.concat(
+            firstSchema.fields().stream().map(Field::name),
+            secondSchema.fields().stream().map(Field::name))
+        .filter(name -> !name.equals(ID_FIELD))
+        .sorted()
```
jyemin, line 149: We can squeeze a bit more efficiency out of this by moving the call to `sorted` after the call to `distinct`, as it will halve (at best) the number of elements needed to be sorted

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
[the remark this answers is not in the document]
jyemin: I don't see it yet in the latest patch.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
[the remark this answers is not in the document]
jyemin: Ah, I was confusing this with the sentinel-checking logic in the `combineFieldSchema` method.

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
rozza: Updated test cases to ensure full coverage of all branches

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: Good catch - done :)

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -107,10 +106,95 @@ private static Schema inferSchema(final String fieldPath, final BsonValue bsonVa
...
+    }
+
+    if (firstSchema.type() != Schema.Type.STRUCT || secondSchema.type() != Schema.Type.STRUCT) {
+      LOGGER.debug("Can't combine non-equal schema that are not both structs");
+      return INCOMPATIBLE_SCHEMA_TYPE;
+    }
+
+    SchemaBuilder builder = SchemaBuilder.struct().name(firstSchema.name()).optional();
+
+    // _id field first
```
rozza, line 140: @jyemin Do you think this is needed for nested structs within arrays?

rozza:
@jyemin ready for review :)

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
jyemin: There is no reason for it in nested structs.  There is likely no need for it anywhere.

jyemin: But seems like you can't remove it here without removing it everywhere, otherwise the order will change depending on whether this is one or more than one nested document in the array.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -107,10 +106,95 @@ private static Schema inferSchema(final String fieldPath, final BsonValue bsonVa
...
-        return DEFAULT_INFER_SCHEMA_TYPE;
+        return Schema.OPTIONAL_STRING_SCHEMA;
     }
   }
 
+  private static Schema combineArrayValueSchema(
+      final Schema firstSchema, final Schema secondSchema) {
+    if (isSentinel(firstSchema)) {
+      return secondSchema;
+    } else if (isSentinel(secondSchema)) {
```
jyemin, line 117: Looks like this condition was actually required.  How come?

rozza: Its just a logic shortcut - it will fall through if not added.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: Removed it in favour to natural sort order.

jyemin: Is there any context where this could be considered a backwards breaking change?

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
jyemin: I'm thinking no, but please confirm.  I didn't really understand the problem you were pointing out.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: Low risk but for example the following document would produce different field orders pre / post change: `{_a: 1, _id: 1}`

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
rozza: Take the following document:

```
{outerArray: [{a: 1}, {a: 'foo'}]}
```

The initial change set the schema to be: `{outerArray: [Type.STRING]}`
But its only the `a` field that has a type conflict: `{outerArray: [{a: Type.STRING}]}`

So the recursive change fixed that.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: Happy to revert c6c69a7 then there would be no backwards incompatibility.

jyemin: Annoyingly the `_` character is also after all the capital letters.  I think we should revert just to be safer.

rozza: Ack

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
jyemin: OK, so that means that every struct is compatible with every other struct.

PR #129, opened 2023-01-24
Fix schema inference

rozza:
Fix nested structs recursion.
Ensure tests recurse correctly when evaluating schemas. Added debug schema helper for comparing and visualizing schemas.

KAFKA-349

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
@@ -16,6 +16,8 @@
 
 package com.mongodb.kafka.connect.source.schema;
 
+import static com.mongodb.kafka.connect.source.schema.SchemaDebugHelper.prettyPrintSchemas;
+
 import java.util.Map;
 import java.util.stream.Stream;
 
@@ -50,12 +52,11 @@ public final class BsonDocumentToSchema {
   private static Schema inferArraySchema(final String fieldPath, final BsonArray bsonArray) {
     Schema combinedSchema = SENTINEL_STRING_TYPE;
     for (final BsonValue v : bsonArray) {
-      combinedSchema = combineArrayValueSchema(combinedSchema, inferSchema(fieldPath, v));
+      combinedSchema = combinedSchema(combinedSchema, inferSchema(fieldPath, v));
       if (combinedSchema == INCOMPATIBLE_SCHEMA_TYPE) {
         break;
       }
     }
-    combinedSchema = isSentinel(combinedSchema) ? Schema.OPTIONAL_STRING_SCHEMA : combinedSchema;
     return SchemaBuilder.array(combinedSchema).name(fieldPath).optional().build();
   }
 
@@ -117,8 +118,7 @@ public final class BsonDocumentToSchema {
     }
   }
 
-  private static Schema combineArrayValueSchema(
-      final Schema firstSchema, final Schema secondSchema) {
+  private static Schema combinedSchema(final Schema firstSchema, final Schema secondSchema) {
     if (isSentinel(firstSchema)) {
       return secondSchema;
     } else if (isSentinel(secondSchema)) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -50,12 +52,11 @@ private static Schema inferDocumentSchema(final String fieldPath, final BsonDocu
   private static Schema inferArraySchema(final String fieldPath, final BsonArray bsonArray) {
     Schema combinedSchema = SENTINEL_STRING_TYPE;
     for (final BsonValue v : bsonArray) {
-      combinedSchema = combineArrayValueSchema(combinedSchema, inferSchema(fieldPath, v));
+      combinedSchema = combinedSchema(combinedSchema, inferSchema(fieldPath, v));
```
rozza, line 55: Just renamed - as it combines any schemas.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -177,25 +183,9 @@ private static Schema combineFieldSchema(final Field firstField, final Field sec
...
-            .name(firstField.name())
-            .optional()
-            .build();
-      }
-    } else {
-      LOGGER.debug("Can't combine non-matching fields: {} and {}", firstField, secondField);
-      return INCOMPATIBLE_SCHEMA_TYPE;
     }
+
+    return combinedSchema(firstField.schema(), secondField.schema());
```
rozza, line 188: Much cleaner now for fields.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -117,8 +118,7 @@ private static Schema inferSchema(final String fieldPath, final BsonValue bsonVa
     }
   }
 
-  private static Schema combineArrayValueSchema(
-      final Schema firstSchema, final Schema secondSchema) {
+  private static Schema combinedSchema(final Schema firstSchema, final Schema secondSchema) {
```
rozza, line 121: This is mostly clean up and ensuring full recursion.

src/main/java/com/mongodb/kafka/connect/source/schema/SchemaDebugHelper.java
```diff
@@ -0,0 +1,119 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.kafka.connect.source.schema;
+
+import org.apache.kafka.connect.data.Field;
+import org.apache.kafka.connect.data.Schema;
+
+public final class SchemaDebugHelper {
```
rozza, line 22: This was added just to help make sense of the schemas.

src/test/java/com/mongodb/kafka/connect/source/producer/SchemaAndValueProducerTest.java
```diff
@@ -160,7 +160,7 @@ void testInferSchemaAndValueProducer() {
                 SchemaBuilder.array(
                         SchemaBuilder.struct()
                             .field("a", Schema.OPTIONAL_INT32_SCHEMA)
-                            .name("arrayComplex_a")
```
rozza, line 163: Sigh - the test asserter was also incorrect and didn't recurse correctly else we would have found this error out.

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
```diff
@@ -198,20 +203,53 @@ void testArraysWithStructs() {
...
+                createArray(
+                    "structsWithMixedTypes",
+                    createStruct("structsWithMixedTypes", SIMPLE_STRUCT_STRINGS)))
             .build();
 
     assertSchemaEquals(expected, inferDocumentSchema(bsonDocument));
   }
 
+  @Test
+  void testArraysWithStructsWithStructs() {
```
rozza, line 223: New test case added for this scenario

src/test/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchemaTest.java
```diff
@@ -198,20 +203,53 @@ void testArraysWithStructs() {
     Schema expected =
         SchemaBuilder.struct()
             .name(DEFAULT_FIELD_NAME)
-            .field("structs", createArray("structs", SIMPLE_STRUCT))
-            .field("structsEmpty", createArray("structsEmpty", SIMPLE_STRUCT))
-            .field("structsEmptyFirst", createArray("structsEmptyFirst", SIMPLE_STRUCT))
-            .field("structsNull", createArray("structsNull", SIMPLE_STRUCT))
-            .field("structsNullFirst", createArray("structsNullFirst", SIMPLE_STRUCT))
-            .field("structsOrdering", createArray("structsOrdering", SIMPLE_STRUCT))
+            .field("structs", createArray("structs"))
```
rozza, line 206: Most tests had to also be changed now `assertSchemaEquals` actually asserts correctly :crying_cat_face:

src/test/java/com/mongodb/kafka/connect/source/schema/SchemaUtils.java
```diff
@@ -99,11 +100,11 @@ public static void assertSchemaEquals(final Schema expected, final Schema actual
     } else {
       assertEquals(expectedDefaultValue, actualDefaultValue, "values differ");
     }
-    if (expected.type() == Type.MAP) {
-      assertEquals(expected.keySchema(), actual.keySchema(), "keySchema differs");
-      assertEquals(expected.valueSchema(), actual.valueSchema(), "valueSchema differs");
+
+    if (expected.type() == Type.ARRAY) {
```
rozza, line 104: :scream:

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -50,12 +52,11 @@ private static Schema inferDocumentSchema(final String fieldPath, final BsonDocu
   private static Schema inferArraySchema(final String fieldPath, final BsonArray bsonArray) {
     Schema combinedSchema = SENTINEL_STRING_TYPE;
     for (final BsonValue v : bsonArray) {
-      combinedSchema = combineArrayValueSchema(combinedSchema, inferSchema(fieldPath, v));
+      combinedSchema = combinedSchema(combinedSchema, inferSchema(fieldPath, v));
       if (combinedSchema == INCOMPATIBLE_SCHEMA_TYPE) {
         break;
       }
     }
-    combinedSchema = isSentinel(combinedSchema) ? Schema.OPTIONAL_STRING_SCHEMA : combinedSchema;
```
rozza, line 58: Had to remove this as this loses the sentinel.  

Note: The sentinel is equivalent to Schema.OPTIONAL_STRING_SCHEMA but not the same instance as it and that is what we check against.

rozza:
Very embarrassing - I couldn't get the new tests to fail as expected before applying a fix and then found the test asserter for schemas was also broken. :sob:

jyemin (commented):
It looks good to me but before I go too deep I want you to have a look at the new issue reported by Joe.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -130,38 +130,44 @@ private static Schema combineArrayValueSchema(
...
+            .sorted()
+            .forEach(
+                name ->
+                    structBuilder.field(
+                        name,
+                        combineFieldSchema(firstSchema.field(name), secondSchema.field(name))));
+        return structBuilder.build();
+      default:
+        logIncompatibleSchemas(firstSchema, secondSchema);
+        return INCOMPATIBLE_SCHEMA_TYPE;
```
jyemin, line 169: This branch isn't covered by tests.  Can you add one?

jyemin (changes requested):

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: We shouldn't hit this code - as the schema types have to be equal see line 132

The default is only there for completeness - it could be converted to an assertion and fail but I opted to log and return a valid type.

jyemin: Could we just remove the conditional on line 132 and then this code will cover the condition?  If so that seems clearer and simpler.

rozza: We have to check both types are the same before switching on the type.  

If `firstSchema` was a Struct and `secondSchema` was a int - then it would fail.

jyemin: Ah ok.  Can you just add a comment like `// Effectively unreachable`

jyemin (approved):
LGTM

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: Will do

PR #130, opened 2023-01-25
Logging improvements when types are in compatible

rozza:
I've updated the logging output to try and improve the debug-ability of inferring schema.

Example:

```
[2023-01-25 14:09:21,025] DEBUG Incompatible Schemas: STRUCT : STRUCT
Schema one:
  | a: int32 (optional = true) (name = null)
  | b: int32 (optional = true) (name = null)
  | c: int32 (optional = true) (name = null)
  | d: int32 (optional = true) (name = null)
  | e: int32 (optional = true) (name = null)
Schema two:
  | a: string (optional = true) (name = null)
  | b: string (optional = true) (name = null)
  | c: string (optional = true) (name = null)
  | d: string (optional = true) (name = null)
  | e: string (optional = true) (name = null)
 (com.mongodb.kafka.connect.source.schema.BsonDocumentToSchema:224)
```

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
@@ -19,6 +19,7 @@ package com.mongodb.kafka.connect.source.schema;
 import static com.mongodb.kafka.connect.source.schema.SchemaDebugHelper.prettyPrintSchemas;
 
 import java.util.Map;
+import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 import org.apache.kafka.connect.data.Decimal;
@@ -37,7 +38,8 @@ public final class BsonDocumentToSchema {
   public static final String DEFAULT_FIELD_NAME = "default";
   private static final Logger LOGGER = LoggerFactory.getLogger(BsonDocumentToSchema.class);
   private static final String ID_FIELD = "_id";
-  static final Schema INCOMPATIBLE_SCHEMA_TYPE = Schema.OPTIONAL_STRING_SCHEMA;
+  static final Schema INCOMPATIBLE_SCHEMA_TYPE =
+      SchemaBuilder.type(Schema.Type.STRING).optional().build();
   static final Schema SENTINEL_STRING_TYPE =
       SchemaBuilder.type(Schema.Type.STRING).optional().build();
 
@@ -130,7 +132,6 @@ public final class BsonDocumentToSchema {
     }
 
     if (firstSchema.type() != secondSchema.type()) {
-      logIncompatibleSchemas(firstSchema, secondSchema);
       return INCOMPATIBLE_SCHEMA_TYPE;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -37,7 +38,8 @@ public final class BsonDocumentToSchema {
   public static final String DEFAULT_FIELD_NAME = "default";
   private static final Logger LOGGER = LoggerFactory.getLogger(BsonDocumentToSchema.class);
   private static final String ID_FIELD = "_id";
-  static final Schema INCOMPATIBLE_SCHEMA_TYPE = Schema.OPTIONAL_STRING_SCHEMA;
+  static final Schema INCOMPATIBLE_SCHEMA_TYPE =
```
jyemin, line 41: Is this now effectively serving as a sentinel and used with reference equality semantics?

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -141,10 +142,13 @@ private static Schema combinedSchema(final Schema firstSchema, final Schema seco
                     combinedSchema(firstSchema.valueSchema(), secondSchema.valueSchema()))
                 .name(firstSchema.name())
                 .optional();
+        logIncompatibleSchemasIf(
```
jyemin, line 145: What is the effect of moving the location of the logging around like this?  I can't visualize it.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -203,7 +214,14 @@ private static String createFieldPath(final String fieldPath, final String field
 
   private static void logIncompatibleSchemas(final Schema firstSchema, final Schema secondSchema) {
     if (LOGGER.isDebugEnabled()) {
```
jyemin, line 216: Also would need to change to `isInfoEnabled` (see below comment)

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
```diff
@@ -163,6 +167,13 @@ private static Schema combinedSchema(final Schema firstSchema, final Schema seco
                     structBuilder.field(
                         name,
                         combineFieldSchema(firstSchema.field(name), secondSchema.field(name))));
+
+        logIncompatibleSchemasIf(
+            () ->
+                structBuilder.fields().stream()
```
jyemin, line 173: This seems like an expensive check to do when logging is enabled, in the case where the it evaluates to false.  But I can't think of a way to avoid it and still get the effect you're looking for.

I'm also wondering how much logging could happen for the same struct.  Say we have an array with 20 documents who's schema need to be combined.  Will this log once for every attempt to combine (i.e. 19 times)?

jyemin (changes requested):

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: Correct - it allows logging of the whole array / document rather than just the lowest individual values.

src/main/java/com/mongodb/kafka/connect/source/schema/BsonDocumentToSchema.java
rozza: This outputs the whole array rather than just the individual incompatible values.

rozza:
Closing - I'll add a ticket for this - so we can prioritize the work. 

I have committed the small change from `SchemaDebugHelper` into master to reduce the verboseness.

PR #136, opened 2023-03-29
KAFKA-360 add config variable to use document id as key on delete events

gonpinho:
Introduces a config variable `document.id.on.tombstone.key` to enable the use of the document id instead of the ChangeStream ObjectId on delete events keys. https://jira.mongodb.org/browse/KAFKA-360


gonpinho:
LGTM @rozza. Thx

PR #140, opened 2023-07-14
Improve "Could not convert value ..." messages

stIncMale:
A viewer should not be left guessing whether the value was a null reference, or a `"null"` `CharSequence`.

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
@@ -32,6 +32,8 @@ import org.apache.kafka.connect.sink.SinkRecord;
 import org.bson.BsonDocument;
 import org.bson.BsonValue;
 
+import com.mongodb.lang.Nullable;
+
 public class LazyBsonDocument extends BsonDocument {
   private static final long serialVersionUID = 1L;
 
@@ -158,7 +160,10 @@ public class LazyBsonDocument extends BsonDocument {
             unwrapped = converter.apply(record.keySchema(), record.key());
           } catch (Exception e) {
             throw new DataException(
-                format("Could not convert key `%s` into a BsonDocument.", record.key()), e);
+                format(
+                    "Could not convert key %s into a BsonDocument.",
+                    unambiguousToString(record.key())),
+                e);
           }
           break;
         case VALUE:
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
```diff
@@ -185,4 +193,15 @@ private Object writeReplace() {
...
   }
+
+  private static String unambiguousToString(@Nullable final Object v) {
+    String vToString = String.valueOf(v);
+    if (v == null) {
+      return format("'%s' (null reference)", vToString);
+    } else if (vToString.equals(String.valueOf((Object) null))) {
+      return format("'%s' (%s, not a null reference)", vToString, v.getClass());
+    } else {
+      return format("'%s' (%s)", vToString, v.getClass());
```
rozza, line 204: Recommend: `v.getClass().getSimpleName()` or `v.getClass().getName()`

rozza (commented):
One recommendation then LGTM

src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
stIncMale: 1. `getSimpleName` erases the package data and returns an empty string for anonymous classes. Given this, `toString`/`getName` is much better for this use case.
2. The only difference between `getName` and `toString` is that `toString` additionally specifies whether it's an interface or a class. Why do you think not having that information is preferable?

rozza: It's only a recommendation. 

Only the following types are supported:

 - Schema'd data - `Struct`
 - Json data - `Map<String, Object>`
 - Raw json string - `String`
 - Raw bytes = `byte[]`

So there are no cases where anonymous classes could be used and or where package data is particularly helpful. Simple and clear error messages help users, too verbose or complex ones can hinder users.

rozza (approved):

src/main/java/com/mongodb/kafka/connect/sink/converter/LazyBsonDocument.java
stIncMale: Used `Class.getName`.

PR #82, opened 2021-08-18
Add regular expressions in topic mapping

JeanFrancoisGuena:
Hi,

I wrote this PR to give users the ability to map change stream messages to topics through regular expressions.

Say that you want all messages from databases starting with "myDb-" mapped to a unique topic, you just have to provide a key-value pair in 'topic.namespace.map' JSON object like this: `{"/myDb-.*/": "myUniqueTopic", ...}`

You may be more specific with regular expressions of the form `"/myDb-.*\\.myColl-.*/"` (it looks like a namespace) that take precedence over `"/myDb-.*/"` ones for example. Notice that these mapping come **after** the ones defined without regex, but **before** the "*" wildcard...

Additionally, I introduced `<database>` and `<collection>` reserved words the user may use to add respectively database name and collection name to the topic. For example,  `{"/myDb-.*/": "myDbTopics.<collection>", ...}` would map messages from databases starting with "myDb-" to a topic named `myDbTopics.<the actual collection name picked from change stream document>`.

I did this to try and stay as compliant as possible with `DefaultTopicMapper` behavior...

Every regular expression must begin and end with the slash character, otherwise we wouldn't know they are regular expressions.

Use of regular expressions requires that default topic mapper be a new `EnrichedTopicMapper` class. So, user has to provide the `topic.mapper=com.mongodb.kafka.connect.source.topic.mapping.EnrichedTopicMapper` configuration.

I did it, not to modify `DefaultTopicMapper` code, however I moved `DefaultTopicMapper` code to a new `AbstractTopicMapper` abstract class, that both `DefaultTopicMapper` and `EnrichedTopicMapper` extend. This way, more topic mapper classes could easily be added in the future...?

I updated `MongoSourceConfig` accordingly, but I do not know where to update the official documentation :-(

Of course, I added unit tests, even if I confess more could be added...

Unfortunately, I was not able to create any JIRA feature request before I submit this PR.

Happy reviewing :-)

PR #141, opened 2023-07-20
Add regular expressions in topic mapping

stIncMale:
The first three commits are purposefully organized to facilitate reviewing. I highly recommend **reviewing them sequentially one by one.** Do not worry, each next commit does not through away the documentation/code added/changed in previous commits, i.e., when reviewing the first three commits sequentially **you won't review something that is deleted later**.

The approach taken in this PR is inspired by https://github.com/mongodb/mongo-kafka/pull/82.

KAFKA-252


src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -184,15 +184,70 @@ public class MongoSourceConfig extends AbstractConfig {
...
+          + "\nThe name in each pair is a namespace pattern, the value is a string representing the corresponding topic name template."
+          + " Pairs are ordered. When there are multiple pairs with equal namespace patterns (duplicates),"
+          + " they are deduplicated and only one pair is taken into account:"
+          + " its position is taken from the first pair among duplicates,"
+          + " its topic name template is taken from the last pair among duplicates."
+          + " After deduplication, pairs with an empty topic name template are ignored when computing topic names."
+          + " Note that a topic name computed based on this configuration is then decorated"
+          + " using the `topic.prefix` and `topic.suffix` configuration properties."
+          + "\nThere are two kinds of pairs:"
+          + "\n- Simple. The namespace pattern must not contain solidus '/' and can be either of the following:"
```
stIncMale, line 199: > Simple. The namespace pattern must not contain solidus '/'

The justification for introducing such a limitation is here: https://jira.mongodb.org/browse/KAFKA-252?focusedCommentId=5572462&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5572462.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -184,15 +184,70 @@ public class MongoSourceConfig extends AbstractConfig {
...
+          + "\n  - A namespace with a collection name, in which case it matches only that namespace."
+          + " The topic name template is interpreted as the topic name."
+          + "\n  - A namespace with only a database name, in which case it matches any namespace having that database name."
+          + " The matching namespace may either have a collection name, or not:"
+          + "\n    - If there is a collection name, then the topic name is computed"
+          + " by concatenating the topic name template and the collection name from the matching namespace, separated by `topic.separator`."
+          + "\n    - If there is no collection name, then the the topic name template is interpreted as the topic name."
+          + "\n- Regex. The namespace pattern starts with solidus '/',"
```
stIncMale, line 207: > Regex. The namespace pattern starts with solidus '/'

Note that if in the future we need to either customize regex pairs, or add a new kind of pairs, we will be able to do that by inventing syntax for the text before the starting '/', and that will be a backward-compatible change.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -184,15 +184,70 @@ public class MongoSourceConfig extends AbstractConfig {
...
-          + "Note: Any prefix and suffix configuration will still apply.";
+      "A JSON object configuring how to map a MongoDB change stream document namespace"
+          + "(a database name optionally concatenated with a collection name, separated by full stop '.')"
+          + " to a Kafka topic name. Used by the `DefaultTopicMapper`."
+          + "\nThe name in each pair is a namespace pattern, the value is a string representing the corresponding topic name template."
+          + " Pairs are ordered. When there are multiple pairs with equal namespace patterns (duplicates),"
+          + " they are deduplicated and only one pair is taken into account:"
+          + " its position is taken from the first pair among duplicates,"
+          + " its topic name template is taken from the last pair among duplicates."
+          + " After deduplication, pairs with an empty topic name template are ignored when computing topic names."
```
stIncMale, line 195: `DefaultTopicMapper` allows duplicates and empty topic name templates. Since we cannot forbid those now, I decided that it is good to have the behavior specified.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -184,15 +184,70 @@ public class MongoSourceConfig extends AbstractConfig {
...
+          + "\n  - `db` The database name from the matching namespace."
+          + "\n  - `sep` The value of the `topic.separator` configuration property."
+          + "\n  - `coll` The collection name from the matching namespace, or an empty string if there is no collection name."
+          + "\n  - `sep_coll` The value of the `coll` variable"
+          + " prefixed with the value of `sep` if and only if the value of `coll` is not empty."
+          + "\n  - `coll_sep` The value of the `coll` variable"
+          + " suffixed with the value of `sep` if and only if the value of `coll` is not empty."
+          + "\n  - `sep_coll_sep` The value of the `coll` variable"
+          + " prefixed and suffixed with the value of `sep` if and only if the value of `coll` is not empty."
+          + "\n To be expanded, a variable must be enclosed between curly brackets '{' and '}', for example '{db}'."
```
stIncMale, line 220: The variable expansion syntax was chosen to be similar to the one proposed in [JEP 430: String Templates (Preview)](https://openjdk.org/jeps/430) with the difference that we don't use `\`.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
```diff
@@ -184,15 +184,70 @@ public class MongoSourceConfig extends AbstractConfig {
...
-      "A json map that maps change stream document namespaces to topics.\n"
-          + "For example: `{\"db\": \"dbTopic\", \"db.coll\": \"dbCollTopic\"}` will map all "
-          + "change stream documents from the `db` database to `dbTopic.<collectionName>` apart from"
-          + "any documents from the `db.coll` namespace which map to the `dbCollTopic` topic.\n"
-          + "If you want to map all messages to a single topic use `*`: "
-          + "For example: `{\"*\": \"everyThingTopic\", \"db.coll\": \"exceptionToTheRuleTopic\"}` "
-          + "will map all change stream documents to the `everyThingTopic` apart from the `db.coll` "
-          + "messages."
-          + "Note: Any prefix and suffix configuration will still apply.";
+      "A JSON object configuring how to map a MongoDB change stream document namespace"
```
stIncMale, line 187: The current online documentation for `topic.namespace.map` is here: https://www.mongodb.com/docs/kafka-connector/current/source-connector/configuration-properties/kafka-topic/.

src/main/java/com/mongodb/kafka/connect/source/topic/mapping/DefaultTopicMapper.java
```diff
@@ -128,4 +192,189 @@ private String undecoratedTopicName(
...
+      throw configExceptionSupplier.apply(e.getMessage());
+    }
+  }
+
+  @FunctionalInterface
+  private interface ConfigExceptionSupplier extends Function<String, ConnectConfigException> {}
+
+  private static final class TopicNameTemplate {
+    private static final char START_VAR_NAME = '{';
+    private static final char END_VAR_NAME = '}';
```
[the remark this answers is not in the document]
rozza: Ack

src/main/java/com/mongodb/kafka/connect/source/topic/mapping/DefaultTopicMapper.java
```diff
@@ -128,4 +198,198 @@ private String undecoratedTopicName(
...
+      return Pattern.compile(regex).matcher("");
+    } catch (PatternSyntaxException e) {
+      throw configExceptionSupplier.apply(e.getMessage());
+    }
+  }
+
+  @FunctionalInterface
+  private interface ConfigExceptionSupplier extends Function<String, ConnectConfigException> {}
+
+  private static final class TopicNameTemplate {
```
rozza, line 215: Nice - Shame had to build all this but it is powerful.

src/test/java/com/mongodb/kafka/connect/source/topic/mapping/DefaultTopicMapperTest.java
```diff
@@ -42,106 +41,110 @@
...
-      BsonDocument.parse("{ns: {db: 'db1', coll: 'coll2'}}");
-
-  private static final String TOPIC_NAMESPACE_MAP =
-      "{\"db1\": \"mappedDBTopic\", \"db1.coll1\": \"mappedDBAndCollTopic\"}";
-
-  private static final String TOPIC_NAMESPACE_ALL_MAP =
-      "{\"*\": \"allTopic\", \"db2.coll2\": \"allExceptionTopic\"}";
+  private static final String TOPIC_NAMESPACE_MAP_EXAMPLE1 =
+      "{'myDb': 'topicTwo', 'myDb.myColl': 'topicOne'}";
+  private static final String TOPIC_NAMESPACE_MAP_EXAMPLE2 =
```
rozza, line 53: Lets test all the template types to ensure they work as expected and help prevent any future regressions.

rozza (changes requested):
Looks really good and seems more powerful than the original PR with the templating. However, the cost is that adds a lot of complexity to the code. I'd advise double checking that the separator templating functionality is needed by end users. Added a request to add tests for all separators.

The main issue I have is its unclear from the ticket and the relative lack of requirements what is required. There is a cryptic comment regarding extra requirements but its unclear as to the context 🤷 

So my question would be is there a requirement where a user wants to use regex to map multiple collections to a single topic? Is that possible?

`{"/(d+)_conversation(.*)": "recentConversations"}`

I also noticed the https://docs.confluent.io/platform/current/connect/transforms/topicregexrouter.html which use regex replacements for changing the end topic. Are there any requirements to do similar?

src/main/java/com/mongodb/kafka/connect/source/topic/mapping/DefaultTopicMapper.java
```diff
@@ -67,12 +69,21 @@ public void configure(final MongoSourceConfig config) {
...
+   *     non-{@linkplain String#isEmpty() empty} MongoDB database name, the method returns an
+   *     {@linkplain String#isEmpty() empty} string.
+   */
   @Override
   public String getTopic(final BsonDocument changeStreamDocument) {
 
     String dbName = getStringFromPath(DB_FIELD_PATH, changeStreamDocument);
     if (dbName.isEmpty()) {
-      return dbName;
+      return "";
```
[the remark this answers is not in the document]
stIncMale: `new String("")` creates a new empty string, but `""` simply refers to an empty string. [JLS 3.10.5. String Literals](https://docs.oracle.com/javase/specs/jls/se20/html/jls-3.html#jls-3.10.5): _"a string literal always refers to the **same** instance of class `String`"_.

The method here returns an empty string with accordance to its specification, and it was returning an empty string previously. I just wanted to make it clear that what the method returns is not a DB name.

src/main/java/com/mongodb/kafka/connect/source/MongoSourceConfig.java
stIncMale: Yeah, "solidus" is the name of this character in Unicode, see [Basic Latin (ASCII)](https://unicode.org/charts/PDF/U0000.pdf) in https://unicode.org/charts/. "slash", "forward slash", "virgule" are aliases.

src/main/java/com/mongodb/kafka/connect/source/topic/mapping/DefaultTopicMapper.java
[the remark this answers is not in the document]
rozza: Ack

stIncMale:
> seems more powerful than the original PR with the templating
> I'd advise double checking that the separator templating functionality is needed by end users.

The difference from the original PR is in the syntax and that we support more variables (the original PR supported only a variable for the DB name and the collection name), but not in the presence of variable expansion (the original implementation of the variable expansion neither detected configuration errors nor were extensible, but it still was there). Supporting specifically the separator variable and its variations does not really add complexity because the variable expansion mechanism is needed for other variables anyway.

I think, the original author simply did not think the behavior around the separator through, which is why he had nothing about it in the PR.

> So my question would be is there a requirement where a user wants to use regex to map multiple collections to a single topic? Is that possible?

That is, of course, possible: a single regular expression namespace pattern may match multiple namespaces, depending on how it is written.

> I also noticed the https://docs.confluent.io/platform/current/connect/transforms/topicregexrouter.html which use regex replacements for changing the end topic. Are there any requirements to do similar?

As you pointed out, there are no clearly formulated requirements, Robert knows this and scheduled the ticket as is. Based on this fact, I understand that there is no need to clarify  the requirements, they are roughly what was proposed in the original PR. There is also no one to clarify them with, the original PR author is [_"on retirement now"_](https://github.com/mongodb/mongo-kafka/pull/82#issuecomment-1160162125). Nonetheless, I [asked Robert](https://jira.mongodb.org/browse/KAFKA-252?focusedCommentId=5572092&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5572092) and got no response.


I am no sure why we would want to duplicate that functionality in our `topic.namespace.map`, given that it already exists as an SMT. But it is doable in the future if needed:

- The syntax that [`Matcher`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html) uses when replacing matched subsequences to refer to captured groups is `$g` and `${gName}` (the latter is for named groups). Neither conflicts with our syntax `{vName}` for variable expansion.
- If we add support for replacing matched subsequences in the future, users will have to enable it by specifying something before the first `/` (we will have to think that part through). If we support this now, nothing special before the first `/` is needed.

src/test/java/com/mongodb/kafka/connect/source/topic/mapping/DefaultTopicMapperTest.java
stIncMale: Added test cases for `EXTENSIVE_TOPIC_NAMESPACE_MAP`.

rozza: Suggest adding cases for: `{sep_coll_sep}` and `{coll_sep}`

rozza (approved):
I believe from the ticket this meets all the requirements.

One suggestion for two more test cases but ultimately LGTM

src/test/java/com/mongodb/kafka/connect/source/topic/mapping/DefaultTopicMapperTest.java
stIncMale: Added for `sep_coll`, `coll_sep`, `sep_coll_sep`.

PR #145, opened 2023-08-09
Source: Added configuration to use the document key for the sourceRec…

rozza:
…ord key

This is potentially a breaking change as the newly added configuration `change.stream.document.key.as.key` defaults to true.

Previously, the resume token was used as the source key, but it limits the usefulness of tombstones both for topic compactions and for downstream implementations. As such this is considered both an improvement and a bug fix.

Set to false to revert back to the previous behaviour.

KAFKA-360


katcharov (approved):
LGTM

PR #147, opened 2023-12-05
Logging incompatible properties no longer NPEs on null values

rozza:
KAFKA-390

----

When performing REST call on `/connector-plugins/com.mongodb.kafka.connect.MongoSinkConnector/config/validate` it appears that some configs can have a `null` value. Use of the stream API caused a NPE, so this change uses `Optional.ofNullable` to work around possible `null` values.

### DIFF
--- a/src/test/java/com/mongodb/kafka/connect/util/config/ConfigSoftValidatorTest.java
+++ b/src/test/java/com/mongodb/kafka/connect/util/config/ConfigSoftValidatorTest.java
@@ -19,7 +19,6 @@ import static com.mongodb.kafka.connect.sink.MongoSinkConfig.TOPIC_OVERRIDE_CONF
 import static java.lang.String.format;
 import static java.util.stream.Collectors.toSet;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


src/test/java/com/mongodb/kafka/connect/util/config/ConfigSoftValidatorTest.java
```diff
@@ -108,6 +109,27 @@ void logIncompatibleProperties() {
...
+        Stream.of(latterIgnoredMsg("a", topicOverridePropertyName("t", "c"))).collect(toSet());
+    Set<String> actualMessages = new HashSet<>();
+    ConfigSoftValidator.logIncompatibleProperties(incompatibleConfigs, props, actualMessages::add);
+    assertEquals(expectedMessages, actualMessages);
+
+    props.put(topicOverridePropertyName("t", "c"), null);
+    actualMessages = new HashSet<>();
+    ConfigSoftValidator.logIncompatibleProperties(incompatibleConfigs, props, actualMessages::add);
+    assertTrue(actualMessages.isEmpty());
+  }
```
stIncMale, line 131: Instead of adding a new test, I propose to modify the existing one. That way the amount of test code remains roughly unchanged (the modified test still fails when run against the current `master` due to each of the new `null` values). See https://github.com/stIncMale/mongo-kafka/commit/935410d412b471f3871c13b7295805b2e51f5261.

stIncMale (changes requested):

vbabanin (approved):
LGTM!

src/test/java/com/mongodb/kafka/connect/util/config/ConfigSoftValidatorTest.java
rozza: Thanks - incorporated your change

PR #148, opened 2023-12-05
Add wildcard support for starts with field name matching

rozza:
KAFKA-391

----

This sink connector has a post processor that allows for field name matching.
(https://www.mongodb.com/docs/kafka-connector/current/sink-connector/fundamentals/post-processors/#projection-wildcard-pattern-matching-examples)

The documentation states:

> Projection Wildcard Pattern Matching Examples
> This section shows how you can configure the projector post processors to match wildcard patterns to match field names.
> 
> Pattern   Description
>   \*  Matches any number of characters in the current level.
>   **  Matches any characters in the current level and all nested levels.

The issue is while matching against fields with subdocuments would work eg: `wind_speed*.average` would match against: `{wind_speed_10ms: {average: x}}`. Just using `wind_speed*` would not match against: `{wind_speed_10ms: x}`.

Note:

This PR **_only implements_** partial field name matching at the top level where the `*` is at the end of the field name. So `wind*speed_10ms` would not be expected to match any fields.

A separate ticket will handle clarifying the limitations in the documentation.

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/processor/field/projection/BlockListProjector.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/processor/field/projection/BlockListProjector.java
@@ -40,14 +40,7 @@ public abstract class BlockListProjector extends FieldProjector {
   }
 
   private void doProjection(final String field, final BsonDocument doc) {
-    if (!field.contains(FieldProjector.SUB_FIELD_DOT_SEPARATOR)) {
-      if (field.equals(FieldProjector.SINGLE_WILDCARD)
-          || field.equals(FieldProjector.DOUBLE_WILDCARD)) {
-        handleWildcard(field, "", doc);
-        return;
-      }
-
-      doc.remove(field);
+    if (processNonNestedMatches(field, doc)) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


stIncMale:
I think, the "into" branch should be `master`, not `KAFKA-390`, shouldn't it?

src/main/java/com/mongodb/kafka/connect/sink/processor/field/projection/BlockListProjector.java
```diff
@@ -40,14 +40,7 @@ protected void projectDocument(final BsonDocument doc) {
...
   private void doProjection(final String field, final BsonDocument doc) {
-    if (!field.contains(FieldProjector.SUB_FIELD_DOT_SEPARATOR)) {
-      if (field.equals(FieldProjector.SINGLE_WILDCARD)
-          || field.equals(FieldProjector.DOUBLE_WILDCARD)) {
-        handleWildcard(field, "", doc);
-        return;
-      }
-
-      doc.remove(field);
+    if (processNonNestedMatches(field, doc)) {
```
vbabanin, line 43: [nit] I usually find that having multiple return statements increases cognitive complexity. It seems that processNonNestedMatches returns false only when !field.contains(FieldProjector.SUB_FIELD_DOT_SEPARATOR) is false. If we move this check before entering the method, we might slightly simplify the code by avoiding return statements inside the method.

if (!field.contains(FieldProjector.SUB_FIELD_DOT_SEPARATOR)) {
        processNonNestedMatches(field, doc);
    }

rozza:
Thanks merged to master

src/main/java/com/mongodb/kafka/connect/sink/processor/field/projection/BlockListProjector.java
rozza: Done

PR #149, opened 2024-01-07
KAFKA-395

tlsdnwn55:
https://jira.mongodb.org/browse/KAFKA-395

Refactored getDeleteWriteModelStrategy method to correctly handle the case when DELETE_WRITEMODEL_STRATEGY_CONFIG is set to DELETE_WRITEMODEL_STRATEGY_DEFAULT. 
Now, the method returns a new DeleteOneDefaultStrategy instance only if the IdStrategy is an instance of FullKeyStrategy, PartialKeyStrategy, or ProvidedInKeyStrategy. 
This ensures proper behavior for the specified configuration.

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
+++ b/src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
@@ -525,22 +525,22 @@ public class MongoSinkTopicConfig extends AbstractConfig {
       return Optional.empty();
     }
     if (deleteOneWriteModelStrategy == null) {
-      if (!getString(DELETE_WRITEMODEL_STRATEGY_CONFIG).isEmpty()) {
-        deleteOneWriteModelStrategy =
-            getWriteModelStrategyFromConfig(DELETE_WRITEMODEL_STRATEGY_CONFIG);
-      } else {
-        /*
-        NOTE: DeleteOneModel requires the key document which means that the only reasonable ID generation strategies are those
-        which refer to/operate on the key document. Thus currently this means the IdStrategy must be either:
+      if (DELETE_WRITEMODEL_STRATEGY_DEFAULT.equals(getString(DELETE_WRITEMODEL_STRATEGY_CONFIG))) {
 
-        FullKeyStrategy
-        PartialKeyStrategy
-        ProvidedInKeyStrategy
-        */
         IdStrategy idStrategy = getIdStrategy();
-        if (!(idStrategy instanceof FullKeyStrategy)
-            && !(idStrategy instanceof PartialKeyStrategy)
-            && !(idStrategy instanceof ProvidedInKeyStrategy)) {
+        if (idStrategy instanceof FullKeyStrategy
+            || idStrategy instanceof PartialKeyStrategy
+            || idStrategy instanceof ProvidedInKeyStrategy) {
+          deleteOneWriteModelStrategy = new DeleteOneDefaultStrategy(idStrategy);
+        } else {
+          /*
+          NOTE: DeleteOneModel requires the key document which means that the only reasonable ID generation strategies are those
+          which refer to/operate on the key document. Thus currently this means the IdStrategy must be either:
+
+          FullKeyStrategy
+          PartialKeyStrategy
+          ProvidedInKeyStrategy
+            */
           throw new ConnectConfigException(
               DELETE_ON_NULL_VALUES_CONFIG,
               getBoolean(DELETE_ON_NULL_VALUES_CONFIG),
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
```diff
@@ -551,7 +551,10 @@ public Optional<WriteModelStrategy> getDeleteWriteModelStrategy() {
                   PartialKeyStrategy.class.getSimpleName(),
                   ProvidedInKeyStrategy.class.getSimpleName()));
         }
-        deleteOneWriteModelStrategy = new DeleteOneDefaultStrategy(idStrategy);
+      } else if (!getString(DELETE_WRITEMODEL_STRATEGY_CONFIG).isEmpty()) {
```
rozza, line 554: This can just be an else as the validator will ensure its not an empty string.

src/test/java/com/mongodb/kafka/connect/sink/MongoSinkConfigTest.java
```diff
@@ -833,6 +833,26 @@ Collection<DynamicTest> testGetSingleValidDeleteWriteModelStrategy() {
...
   }
 
+  @Test
+  @DisplayName("test Default DELETE_WRITEMODEL_STRATEGY_CONFIG")
+  void testDeletStrategyDefaultConfig() {
+    Map<String, String> map = createConfigMap();
+    map.put(DELETE_ON_NULL_VALUES_CONFIG, "true");
+    map.put(DOCUMENT_ID_STRATEGY_CONFIG, FullKeyStrategy.class.getName());
+    MongoSinkConfig cfg = new MongoSinkConfig(map);
+    /*
```
rozza, line 843: Can remove this comment.

rozza (changes requested):
Looks good -just a couple of minor suggestions.

src/main/java/com/mongodb/kafka/connect/sink/MongoSinkTopicConfig.java
tlsdnwn55: you're right
If it doesn't fit FULLY_QUALIFIED_CLASS_NAME it will fail.
However, if the user does not set delete.write.model, the default value is applied and the appropriate genetaredId method according to the id strategy is used.

src/test/java/com/mongodb/kafka/connect/sink/MongoSinkConfigTest.java
tlsdnwn55: ok i got it

tlsdnwn55:
I have additional questions.
To tell you in advance, I don't know much about Mongodb.

I am configuring a pipeline using confleunt.
The source of Mongodb is mysql.
Currently, mysql has configured CDC using debezium mysql and is only receiving after images using ExtractNewRecordState. (Notice in advance that the cdc handler option cannot be used.)

As you know, mysql does not have _id Key, which is Mongodb's basic unique index.
To handle the tombstone event, PartialKeyStrategy was applied and the field corresponding to the pk of the source was specified.
And I created a unique index in Mongodb's target collection with the fields set above.
Currently, Insert and Update work well, but the Delete model does not work even though I changed the code as I modified it.
And I found that even if I create deleteFilter with PartialKeyStrategy's generatedId method, it creates BsonDocument with _id as Key.

for example,
The key value for the order topic is order_id, which is pk.
What I want db.collection.deleteOne({"order_id" : 111}) 
but No matter what id strategy I use, deleteFilter doesn't work
It was created as {"_id" : {{"order_id" : 111}} and could not be deleted.

On the other hand, I confirmed that deleteone works well if I only remove _id for a collection where order_id is a unique index.
As I said, in my environment, the source of Mongodb is mysql, so it is impossible to know the _id value of data deleted from mysql.

Can you give me some advice on this?
-link
[DeleteOneDefaultStrategy](https://github.com/mongodb/mongo-kafka/blob/master/src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneDefaultStrategy.java)

rozza:
Merged in [b2e00f4](https://github.com/mongodb/mongo-kafka/commit/b2e00f4a9d59b069bb436449ce2d35cdb30c2c91)

rozza:
Hi @tlsdnwn55,

Thanks for the PR - I have merged for release.

Regarding, the delete handling, I think that may require a custom WriteModelStrategy to encapsulate the logic for all operations.

Ross

tlsdnwn55:
@rozza 
I think that is good idea.
thank you!

PR #158, opened 2024-05-08
Kafka 414

rasifmahmud:
https://jira.mongodb.org/browse/KAFKA-414

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java
+++ b/src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java
@@ -17,6 +17,7 @@ package com.mongodb.kafka.connect.util;
 
 import static java.lang.String.format;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -191,18 +192,17 @@ public final class ConfigHelper {
     return Optional.empty();
   }
 
-  public static Config evaluateConfigValues(
-      final Config rawConfig, final AbstractConfig resolvedConfig) {
-    final Map<String, ?> evalValues = resolvedConfig.values();
-    rawConfig
-        .configValues()
-        .forEach(
-            configValue -> {
-              Object ev = evalValues.get(configValue.name());
-              if (ev != null) {
-                configValue.value(ev);
+  public static Map<String, String> evaluateConfigValues(
+    final Map<String, String> rawConfigs, final AbstractConfig resolvedConfig) {
+    Map<String, String> resolvedRawConfigs = new HashMap<>(rawConfigs);
+    final Map<String, Object> originals = resolvedConfig.originals();
+    rawConfigs
+        .forEach((key, val) -> {
+              Object ev = originals.get(key);
+              if (ev instanceof String) {
+                resolvedRawConfigs.put(key, (String) ev);
               }
             });
-    return rawConfig;
+    return resolvedRawConfigs;
   }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java
```diff
@@ -191,18 +192,17 @@ public static Optional<ConfigValue> getConfigByNameWithoutErrors(
...
-        .configValues()
-        .forEach(
-            configValue -> {
-              Object ev = evalValues.get(configValue.name());
-              if (ev != null) {
-                configValue.value(ev);
-              }
-            });
-    return rawConfig;
+  public static Map<String, String> evaluateConfigValues(
```
jagadishmdb, line 195: This doesnt look right.  Why do we want to get from originals?  I thought values like in the above code is resolved configs which we want to use.

rasifmahmud: Hi @jagadishmdb you are right. In the above code, **configValue.value(ev)** replaces the unresolved value with the resolved value. But the problem is configValue has another List property named "errorMessages". And that gets populated earlier. Please Check: MongoSinkConnector -> validate -> super.validate(connectorConfigs) call. errorCheckingPasswordValueValidator method adds that when super.validate is called with raw unresolved configs. 

So even though configValue.value is replaced with the resolved value, configValue.errorMessages is non empty at this point. Which eventually causes to throw the error down the line.  

Now, let's talk about the below code segment. Even though I am reading from originals, originals property actually has all the resolved values (sounds confusing!). The answer is in the constructor of the AbstractConfig class

`this.originals = resolveConfigVariables(configProviderProps, (Map<String, Object>) originals);` 

So the originals property has all the resolved values. And in my PR I am calling "super.validate" after raw config has been resolved, so the errorCheckingPasswordValueValidator does not get the opportunity to add that error message to the errorMessages property.

jagadishmdb:
Pardon me, I now understand that the issue is calling super.validate before we actually try to resolve the config.   I think we should move the validate for both source and sink.  I do not understand why we need in the evaluate function.

rasifmahmud:
> Pardon me, I now understand that the issue is calling super.validate before we actually try to resolve the config. I think we should move the validate for both source and sink. I do not understand why we need in the evaluate function.

As long as we are calling super.validate on the resolved configs, we should be good. 

I think there can be other ways to transform unresolved configs to resolved configs . What I proposed is just one of them. Feel free to experiment to find a better solution. Thanks

src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java
```diff
@@ -191,18 +192,17 @@ public static Optional<ConfigValue> getConfigByNameWithoutErrors(
...
+      final Map<String, String> rawConfigs, final AbstractConfig resolvedConfig) {
+    Map<String, String> resolvedRawConfigs = new HashMap<>(rawConfigs);
+    final Map<String, Object> originals = resolvedConfig.originals();
```
rasifmahmud, line 198: @jagadishmdb 
The other solution I can think of is change this line #198 to:
`final Map<String, ?> originals = resolvedConfig.values();`
I can do some testing with this tomorrow. I think this is going to be better than getting the originals.

rasifmahmud: I did some analysis with this approach but I think the previous solution was better. values property are typecasted based on ConfigDef. So there is a chance something may be ignored due to type issue in future.

jagadishmdb:
This is added to a new pull request.

PR #160, opened 2024-05-13
Feature/custom auth provider plugpoint

nilaysundarkar:
Implements a plug point for injecting a custom credential provider in the mongo client for sink and source connectors.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -49,7 +49,7 @@ repositories {
 }
 
 extra.apply {
-    set("mongodbDriverVersion", "[4.7,4.7.99)")
+    set("mongodbDriverVersion", "4.9.1")
     set("kafkaVersion", "2.6.0")
     set("avroVersion", "1.9.2")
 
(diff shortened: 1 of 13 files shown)


build.gradle.kts
```diff
@@ -49,7 +49,7 @@ repositories {
 }
 
 extra.apply {
-    set("mongodbDriverVersion", "[4.7,4.7.99)")
+    set("mongodbDriverVersion", "4.9.1")
```
jagadishmdb, line 52: Any specific reason for this change?

nilaysundarkar: Hi @jagadishmdb ,

I made that change thinking that it is needed for MONGODB-AWS auth mechanism. We have been using 4.9.1 drivers so far.

After checking for compatibility at - https://www.mongodb.com/docs/drivers/java/sync/v4.7/fundamentals/auth/#std-label-mongodb-aws-auth-mechanism, I see that 4.7 supports it as well. I am testing with 4.7 to confirm. I think we can keep 'mongodbDriverVersion' to the same value as before, if that is desirable.

I will let you know how the tests go.

Thanks,
Nilay.

nilaysundarkar: Tests successful with the existing mongo driver versions. Please let me know if I should change the PR for restoring mongodbDriverVersion. I see that it has been already changed at https://github.com/mongodb/mongo-kafka/pull/161.

Thanks,
Nilay.

jagadishmdb: Thanks for confirming Nilay!  There are some additional tests we run which were getting different error codes than what we were expecting with 4.9.1 driver.   I will take care of fixing those later.   Thanks for the contribution!

PR #162, opened 2024-05-23
Extract upsert from sink strategies as configurable option

PushUpek:
I extracted upsert from sink strategies to the topic configuration. Because I needed the ability to turn it off.


nicolas-moteau:
Hello,
I'm interested in this feature that allows using upsert=false.
Is there any chance to get it merged ?

PR #164, opened 2024-08-09
Gradle: Update spotless and klint versions

rozza:
Also apply any automatic code style changes by
the latest ktlint.

KAFKA-421


rozza:
Seems a duplicate of #167

PR #166, opened 2024-10-03
bug where copy existing is not persisted after it has finished copying 

rahulbats:
…sk switched to change stream

the isCopying flag tracks wether the connector task is copying collection data or tracking change stream. This flag needs to be persisted, so that if the connector restarts later it does no start copying the collection again. There was a if statement which persisted this flag only when iscopying is true which is wrong, as the iscopying flag needs to be persisted in offsets after it has turned false when it has finished copying


src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
```diff
@@ -215,9 +215,9 @@ private List<SourceRecord> pollInternal() {
             changeStreamDocument -> {
               Map<String, String> sourceOffset = new HashMap<>();
               sourceOffset.put(ID_FIELD, changeStreamDocument.getDocument(ID_FIELD).toJson());
-              if (isCopying) {
-                sourceOffset.put(COPY_KEY, "true");
-              }
+              
+              sourceOffset.put(COPY_KEY,  String.valueOf(isCopying));
```
arahmanan, line 219: @rahulbats thanks for opening a pull request for this! We don't check the value of the `COPY_KEY` element within the source offset. That's the case [here](https://github.com/mongodb/mongo-kafka/blob/6e43b1d97f52be6d862328e442eea8ece5492d30/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java#L231) and [here](https://github.com/mongodb/mongo-kafka/blob/8068a450953da6803fe27070ac1629387ca7e26c/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L686). The **existence** alone of the `COPY_KEY` key signals that the existing data should be copied. It seems like we are handling things properly without these changes. Would you be able to provide more details on the issue you experienced?

mdb-gtenrreiro: @rahulbats, @arahmanan has a good point; shouldn't you also check the value of the COPY_KEY? not just that it exists ? otherwise the connector will copy even if the COPY_KEY value is false, no?

rahulbats: hi @arahmanan you are right it only checks the key exists. I  changed that code in my PR to verify if it is true as well. If we dont do this when the connector restarts it will reread the collection again . The isCopying false has to be persisted in the offsets topic , so that when connector restarts it sees the flag is false now and doesnt start rereading the entire collection again. This has been a big issue with out customer and they have giant collections which are getting reread generating lot of dups and wastage of time

arahmanan: Hi @rahulbats, we would need to fix [this if statement](https://github.com/mongodb/mongo-kafka/blob/8068a450953da6803fe27070ac1629387ca7e26c/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L686) as well or we won't resume from the persisted resume token on restarts.

That being said, I don't see a change in behavior when the connector restarts with these changes. `shouldCopyData` was already returning `false` when the `COPY_KEY` wasn't persisted. The `getResumeToken` function seems to have the correct logic as well (i.e. return a resume token if the `COPY_KEY` is not persisted). Those are the only two places ([here](https://github.com/mongodb/mongo-kafka/blob/6e43b1d97f52be6d862328e442eea8ece5492d30/src/main/java/com/mongodb/kafka/connect/source/MongoSourceTask.java#L231) and [here](https://github.com/mongodb/mongo-kafka/blob/8068a450953da6803fe27070ac1629387ca7e26c/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L686)) where the persisted `COPY_KEY` is used. I believe something else is causing the issue your customer is running into. 
Do you have more details from the customer, such as logs and MongoDB version?

arahmanan: Chatted with Rahul about this. The `COPY_KEY` could be set to true in a prior poll.

arahmanan (approved):
thank you! LGTM!

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
Calvinnix: Can you remove the leading space here as well as the extra new lines on :218 and :220

Also I think the build failures have to do with formatting issues with the if conditions.

Calvinnix: Just to break this down visually it sounds like the kafka offset persists the COPY_KEY because when we update the offset we aren't updating the value.

i.e.
kafka offset
```
{
  ID: 1,
  COPY_KEY: true
}
```

<- no longer copying, updating offset to 2
```
{
  ID: 2
}
```

resulting kafka offset:
```
{
  ID: 2,
  COPY_KEY: true
}
```

?

rahulbats: i dont think this is what happens. there is no 
{
  ID: 2
}
Nothing is written if COPY_KEY is false

rahulbats: yes this looks right

Calvinnix: Quick update, I just put up a [PR](https://github.com/mongodb/mongo-kafka/pull/168) that should address the issue you are seeing.

Calvinnix:
This was fixed in https://github.com/mongodb/mongo-kafka/pull/168

closing this PR

PR #168, opened 2024-10-16
KAFKA-428: Mark the copy complete in the source offset for the last copied document

Calvinnix:
This fixes an issue that was occurring when:
1. startup mode `COPY_EXISTING` is used
2. the copy has completed
3. but no new has been sent
4. then the kafka connector restarts

The resulting behavior of this scenario is that the copy would occur again on restart.

---

This was happening because sourceOffets for copied records specify `copy: true`, so the restart thinks that the copy was in progress and it tries again. (_note that restarts that occur during a copy is expected to reattempt the copy, which will duplicate data_)

Here is an example of what that copy offset looks like:
```
{"_id":"{\"_id\": {\"$oid\": \"670ee1efaa5a9af80d592c47\"}, \"copyingData\": true}","copy":"true"}
```

---

It does look like we have [logic](https://github.com/mongodb/mongo-kafka/blob/1a09960e1f90c7ab077c1f1d6a502d162f88e5fb/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L583-L589) which appears to try to `"mark copying ended"` but this doesn't work when the `cachedResult` is null. 
```
      // Copying finished - mark copying ended and add cached result
      isCopying = false;
      LOGGER.info("Finished copying existing data from the collection(s).");
      if (cachedResult != null) {
        batch.add(cachedResult);
        cachedResult = null;
      }
```
(I'm not actually sure if this logic ever worked, but I could see a future where the `cachedResult` concept is removed. I haven't thought about this enough to form a strong enough opinion yet)

---

The fix for this issue was to identify when we are creating the `sourceOffset` for the last copied document and in that case go ahead and `"mark copying ended"` by removing the `copy: true` flag and setting the '_id' field to the `cachedResumeToken`.

### DIFF
--- a/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
+++ b/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
@@ -548,6 +548,7 @@ public class MongoSourceTaskIntegrationTest extends MongoKafkaTestCase {
   @Test
   @DisplayName("Copy existing with a restart midway through")
   void testCopyingExistingWithARestartMidwayThrough() {
+    assumeTrue(isGreaterThanThreeDotSix());
     try (AutoCloseableSourceTask task = createSourceTask()) {
 
       MongoCollection<Document> coll = getCollection();
@@ -620,6 +621,7 @@ public class MongoSourceTaskIntegrationTest extends MongoKafkaTestCase {
   @Test
   @DisplayName("Copy existing with a restart after finishing")
   void testCopyingExistingWithARestartAfterFinishing() {
+    assumeTrue(isGreaterThanThreeDotSix());
     try (AutoCloseableSourceTask task = createSourceTask()) {
 
       MongoCollection<Document> coll = getCollection();
--- a/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
+++ b/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
@@ -36,6 +36,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertIterableEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -186,14 +187,14 @@ public class MongoSourceTaskIntegrationTest extends MongoKafkaTestCase {
               assertTrue(
                   firstPoll.stream()
                       .map(SourceRecord::sourceOffset)
-                      .limit(149)
+                      .limit(150 - 1) // exclude the last record
                       .allMatch(i -> i.containsKey("copy"))),
           // make sure that the last copied element does not have the "copy" key
           () ->
               assertTrue(
                   firstPoll.stream()
                       .map(SourceRecord::sourceOffset)
-                      .skip(149)
+                      .skip(150 - 1) // exclude the last record
                       .findFirst()
                       .filter(i -> !i.containsKey("copy"))
                       .isPresent()));
@@ -558,7 +559,7 @@ public class MongoSourceTaskIntegrationTest extends MongoKafkaTestCase {
               put(MongoSourceConfig.COLLECTION_CONFIG, coll.getNamespace().getCollectionName());
               put(MongoSourceConfig.STARTUP_MODE_CONFIG, StartupMode.COPY_EXISTING.propertyValue());
               put(MongoSourceConfig.POLL_MAX_BATCH_SIZE_CONFIG, "25");
-              put(MongoSourceConfig.POLL_AWAIT_TIME_MS_CONFIG, "10000");
+              put(MongoSourceConfig.POLL_AWAIT_TIME_MS_CONFIG, "1000");
             }
           };
 
@@ -570,6 +571,17 @@ public class MongoSourceTaskIntegrationTest extends MongoKafkaTestCase {
       assertTrue(
           firstPoll.stream().map(SourceRecord::sourceOffset).allMatch(i -> i.containsKey("copy")));
 
+      Map<String, ?> lastOffset = firstPoll.get(25 - 1).sourceOffset();
+
+      // mock the context so that on restart we know where the last task left off
+      when(context.offsetStorageReader()).thenReturn(offsetStorageReader);
+      assertNotNull(lastOffset.get("_id")); // check to make sure the value is an Object
+      @SuppressWarnings("unchecked")
+      Map<String, Object> mockedOffset = (Map<String, Object>) lastOffset;
+      when(offsetStorageReader.offset(any())).thenReturn(mockedOffset);
+      task.initialize(context);
+
+      // perform a restart
       task.stop();
       task.start(cfg);
 
@@ -584,13 +596,13 @@ public class MongoSourceTaskIntegrationTest extends MongoKafkaTestCase {
       assertTrue(
           thirdPoll.stream()
               .map(SourceRecord::sourceOffset)
-              .limit(24)
+              .limit(25 - 1) // exclude the last record in the batch
               .allMatch(i -> i.containsKey("copy")));
       // Make sure the last copied element does not contain the "copy" key
       assertTrue(
           thirdPoll.stream()
               .map(SourceRecord::sourceOffset)
-              .skip(24)
+              .skip(25 - 1) // exclude the last record in the batch
               .findFirst()
               .filter(i -> !i.containsKey("copy"))
               .isPresent());
@@ -602,6 +614,71 @@ public class MongoSourceTaskIntegrationTest extends MongoKafkaTestCase {
       assertSourceRecordValues(createInserts(51, 75), fourthPoll, coll);
       assertFalse(
           fourthPoll.stream().map(SourceRecord::sourceOffset).anyMatch(i -> i.containsKey("copy")));
+    }
+  }
+
+  @Test
+  @DisplayName("Copy existing with a restart after finishing")
+  void testCopyingExistingWithARestartAfterFinishing() {
+    try (AutoCloseableSourceTask task = createSourceTask()) {
+
+      MongoCollection<Document> coll = getCollection();
+
+      HashMap<String, String> cfg =
+          new HashMap<String, String>() {
+            {
+              put(MongoSourceConfig.DATABASE_CONFIG, coll.getNamespace().getDatabaseName());
+              put(MongoSourceConfig.COLLECTION_CONFIG, coll.getNamespace().getCollectionName());
+              put(MongoSourceConfig.STARTUP_MODE_CONFIG, StartupMode.COPY_EXISTING.propertyValue());
+              put(MongoSourceConfig.POLL_MAX_BATCH_SIZE_CONFIG, "25");
+              put(MongoSourceConfig.POLL_AWAIT_TIME_MS_CONFIG, "1000");
+            }
+          };
+
+      insertMany(rangeClosed(1, 50), coll);
+      task.start(cfg);
+
+      List<SourceRecord> firstPoll = getNextBatch(task);
+      assertSourceRecordValues(createInserts(1, 25), firstPoll, coll);
+      assertTrue(
+          firstPoll.stream().map(SourceRecord::sourceOffset).allMatch(i -> i.containsKey("copy")));
+
+      List<SourceRecord> secondPoll = getNextBatch(task);
+      assertSourceRecordValues(createInserts(26, 50), secondPoll, coll);
+      // Make sure all elements, except the last one, contains the "copy" key
+      assertTrue(
+          secondPoll.stream()
+              .map(SourceRecord::sourceOffset)
+              .limit(25 - 1) // exclude the last record in the batch
+              .allMatch(i -> i.containsKey("copy")));
+
+      Map<String, ?> lastOffset = secondPoll.get(25 - 1).sourceOffset();
+
+      // Make sure the last copied element does not contain the "copy" key
+      assertFalse(lastOffset.containsKey("copy"));
+
+      // mock the context so that on restart we know where the last task left off
+      when(context.offsetStorageReader()).thenReturn(offsetStorageReader);
+      assertNotNull(lastOffset.get("_id")); // check to make sure the value is an Object
+      @SuppressWarnings("unchecked")
+      Map<String, Object> mockedOffset = (Map<String, Object>) lastOffset;
+      when(offsetStorageReader.offset(any())).thenReturn(mockedOffset);
+      task.initialize(context);
+
+      // perform a restart
+      task.stop();
+      task.start(cfg);
+
+      // make sure that a copy doesn't occur again because all data was already copied
+      assertTrue(getNextBatch(task).isEmpty());
+
+      // make sure that we can continue to process data
+      insertMany(rangeClosed(51, 75), coll);
+
+      List<SourceRecord> thirdPoll = getNextBatch(task);
+      assertSourceRecordValues(createInserts(51, 75), thirdPoll, coll);
+      assertFalse(
+          thirdPoll.stream().map(SourceRecord::sourceOffset).anyMatch(i -> i.containsKey("copy")));
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -169,6 +169,9 @@ void testSourceLoadsDataFromMongoClientWithCopyExisting() {
           new HashMap<String, String>() {
             {
               put(MongoSourceConfig.STARTUP_MODE_CONFIG, StartupMode.COPY_EXISTING.propertyValue());
+              String namespaceRegex =
+                  String.format("(%s\\.coll|%s\\.coll)", db1.getName(), db2.getName());
+              put(MongoSourceConfig.COPY_EXISTING_NAMESPACE_REGEX_CONFIG, namespaceRegex);
```
Calvinnix, line 174: [driveby] without this filter this test tries to copy all data

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
```diff
@@ -210,68 +211,80 @@ private List<SourceRecord> pollInternal() {
...
+      // copy is in progress. However, for the last document that we are copying, we should not set
+      // this flag because the copy has completed, otherwise we are relying on future change stream
+      // events to signify that we are no longer copying. We also need to set the _id field to be a
+      // valid resume token, which during copying exists in the cachedResumeToken variable.
+      boolean lastDocument = !batchIterator.hasNext();
+      boolean noMoreDataToCopy = copyDataManager != null && !copyDataManager.isCopying();
+      if (isCopying && lastDocument && noMoreDataToCopy) {
+        sourceOffset.put(ID_FIELD, cachedResumeToken.toJson());
+        sourceOffset.remove(COPY_KEY);
+      }
```
Calvinnix, line 234: Highlighting this since this is the logic that fixes the issue.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -566,8 +580,20 @@ void testCopyingExistingWithARestartMidwayThrough() {
 
       List<SourceRecord> thirdPoll = getNextBatch(task);
       assertSourceRecordValues(createInserts(26, 50), thirdPoll, coll);
+      // Make sure all elements, except the last one, contains the "copy" key
+      assertTrue(
+          thirdPoll.stream()
+              .map(SourceRecord::sourceOffset)
+              .limit(24)
```
arahmanan, line 587: same optional comment here

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
```diff
@@ -210,68 +211,80 @@ private List<SourceRecord> pollInternal() {
...
-    getNextBatch()
-        .forEach(
-            changeStreamDocument -> {
-              Map<String, String> sourceOffset = new HashMap<>();
-              sourceOffset.put(ID_FIELD, changeStreamDocument.getDocument(ID_FIELD).toJson());
-              if (isCopying) {
-                sourceOffset.put(COPY_KEY, "true");
-              }
+    Iterator<BsonDocument> batchIterator = getNextBatch().iterator();
+    BsonDocument changeStreamDocument;
```
arahmanan, line 215: [nit] Do we need to define this outside the while loop?

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -566,8 +580,20 @@ void testCopyingExistingWithARestartMidwayThrough() {
...
+              .allMatch(i -> i.containsKey("copy")));
+      // Make sure the last copied element does not contain the "copy" key
       assertTrue(
-          thirdPoll.stream().map(SourceRecord::sourceOffset).allMatch(i -> i.containsKey("copy")));
+          thirdPoll.stream()
+              .map(SourceRecord::sourceOffset)
+              .skip(24)
+              .findFirst()
+              .filter(i -> !i.containsKey("copy"))
+              .isPresent());
```
arahmanan, line 597: Can we add a test where a restart happens (similar to how it's done [here](https://github.com/mongodb/mongo-kafka/blob/1a09960e1f90c7ab077c1f1d6a502d162f88e5fb/src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java#L574-L575)) after copying all documents and confirming that a second poll won't fetch any data?

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
```diff
@@ -210,68 +211,80 @@ private List<SourceRecord> pollInternal() {
...
-            });
+      // if isCopying is true, we want to set the COPY_KEY flag so that kafka has context that a
+      // copy is in progress. However, for the last document that we are copying, we should not set
+      // this flag because the copy has completed, otherwise we are relying on future change stream
+      // events to signify that we are no longer copying. We also need to set the _id field to be a
+      // valid resume token, which during copying exists in the cachedResumeToken variable.
+      boolean lastDocument = !batchIterator.hasNext();
+      boolean noMoreDataToCopy = copyDataManager != null && !copyDataManager.isCopying();
+      if (isCopying && lastDocument && noMoreDataToCopy) {
+        sourceOffset.put(ID_FIELD, cachedResumeToken.toJson());
```
arahmanan, line 232: great find! Should we add a test that confirms we're setting the expected resume token?

arahmanan (commented):
Looking good! I did a first pass

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
arahmanan, line 232: Instead of `cachedResumeToken `, it might be safer to create and use a different variable here that contains the same value as `cachedResumeToken` but isn't modified anywhere else. `cachedResumeToken` get's set to `null` ([here](https://github.com/mongodb/mongo-kafka/blob/1a09960e1f90c7ab077c1f1d6a502d162f88e5fb/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L697)) within `getResumeToken`, which ends up being called [here](https://github.com/mongodb/mongo-kafka/blob/1a09960e1f90c7ab077c1f1d6a502d162f88e5fb/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L596) if `cursor == null`. The `cursor` could be set to `null` if the cluster runs into a temporary connection issue.

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
Calvinnix: Err nope 😆 I'll update

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
Calvinnix: Yeah that's a good catch, I'll create a new token called `resumeTokenAfterCopy`

Calvinnix: Wait I think we're actually good here, I'm happy to create a new variable for extra safety though if you're still concerned.

because we are assigning this in the copied batch the `cursor == null` code won't be hit until after we use and store the `cachedResumeToken`. i.e. the getNextBatch function exits early [here](https://github.com/mongodb/mongo-kafka/blob/1a09960e1f90c7ab077c1f1d6a502d162f88e5fb/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L581).

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -178,11 +181,22 @@ void testSourceLoadsDataFromMongoClientWithCopyExisting() {
           () -> assertEquals(150, firstPoll.size()),
           () -> assertSourceRecordValues(createInserts(1, 75), firstPoll, coll1),
           () -> assertSourceRecordValues(createInserts(1, 75), firstPoll, coll2),
+          // make sure all elements, except the last, contains the "copy" key
           () ->
               assertTrue(
                   firstPoll.stream()
                       .map(SourceRecord::sourceOffset)
-                      .allMatch(i -> i.containsKey("copy"))));
+                      .limit(149)
```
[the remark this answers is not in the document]
Calvinnix: Hmm what do you think about changing this to `150 - 1` with a comment like `// exclude the last record`

I extracted these numbers into variables but I feel like it hurt the readability, especially for the other test.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
Calvinnix: Hmm its sort of misleading but that test you linked technically doesn't correctly test restarts because these tests aren't referencing a cached state.. i.e. shouldCopyData is always true in this case because `context` which is a magic value that comes from Kafka is always null.

Let me check to make sure this isn't something that is wrong with my environment, potentially I'm missing a setup step.

Calvinnix: Ok I can actually mock this

src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
Calvinnix: I think we got coverage of this with the recent e2e tests I added around restarting after copying. Let me know if you disagree.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -544,7 +559,7 @@ void testCopyingExistingWithARestartMidwayThrough() {
               put(MongoSourceConfig.COLLECTION_CONFIG, coll.getNamespace().getCollectionName());
               put(MongoSourceConfig.STARTUP_MODE_CONFIG, StartupMode.COPY_EXISTING.propertyValue());
               put(MongoSourceConfig.POLL_MAX_BATCH_SIZE_CONFIG, "25");
-              put(MongoSourceConfig.POLL_AWAIT_TIME_MS_CONFIG, "10000");
+              put(MongoSourceConfig.POLL_AWAIT_TIME_MS_CONFIG, "1000");
```
Calvinnix, line 562: [driveby] because this test expects an [empty batch](https://github.com/mongodb/mongo-kafka/pull/168/files#diff-23579344359cd812b6c4b727ba54786d597e1ea660faad4d156930d66e526f3fR610) we should reduce the poll await time even further, otherwise we're guaranteeing to wait for at least 5x10000ms. This just speeds the test up.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -556,6 +571,17 @@ void testCopyingExistingWithARestartMidwayThrough() {
       assertTrue(
           firstPoll.stream().map(SourceRecord::sourceOffset).allMatch(i -> i.containsKey("copy")));
 
+      Map<String, ?> lastOffset = firstPoll.get(25 - 1).sourceOffset();
+
+      // mock the context so that on restart we know where the last task left off
+      when(context.offsetStorageReader()).thenReturn(offsetStorageReader);
+      assertNotNull(lastOffset.get("_id")); // check to make sure the value is an Object
+      @SuppressWarnings("unchecked")
+      Map<String, Object> mockedOffset = (Map<String, Object>) lastOffset;
```
Calvinnix, line 580: A lot of this code is just to make the `(Map<String, Object>) lastOffset` cast safer.. since this is just a test we could probably skip the null check.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -556,6 +571,17 @@ void testCopyingExistingWithARestartMidwayThrough() {
...
+      when(offsetStorageReader.offset(any())).thenReturn(mockedOffset);
```
Calvinnix, line 581: [driveby] I included mocking logic in this test because otherwise the code will always re-copy the data, but not because it sees the expected copy offset, but because the context is null.

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -579,6 +617,71 @@ void testCopyingExistingWithARestartMidwayThrough() {
...
+              .allMatch(i -> i.containsKey("copy")));
+
+      Map<String, ?> lastOffset = secondPoll.get(25 - 1).sourceOffset();
+
+      // Make sure the last copied element does not contain the "copy" key
+      assertFalse(lastOffset.containsKey("copy"));
+
+      // mock the context so that on restart we know where the last task left off
+      when(context.offsetStorageReader()).thenReturn(offsetStorageReader);
+      assertNotNull(lastOffset.get("_id")); // check to make sure the value is an Object
```
Calvinnix, line 662: Same comment as the one above for mocking offset, we could remove this line if we want.

Calvinnix:
So the integration tests are failing for version 3.6 which I'm assuming has something to do with changestream/resumetoken support. I'm wondering if I should just remove this test (and potentially replace with newer versions of mongodb)

cc: @arahmanan

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
```diff
@@ -579,6 +617,71 @@ void testCopyingExistingWithARestartMidwayThrough() {
     }
   }
 
+  @Test
```
arahmanan, line 620: > So the integration tests are failing for version 3.6 which I'm assuming has something to do with changestream/resumetoken support. I'm wondering if I should just remove this test (and potentially replace with newer versions of mongodb)
EDIT: I added some logging in this [patch](https://spruce.mongodb.com/task/mongo_kafka_integration_tests__version~3.6_os~ubuntu_topology~replicaset_javaVersion~JDK17_integration_test_task_patch_d477f00c5506ac461ebc86cbeaaff291598d921c_67116e328067f60007286765_24_10_17_20_06_24?execution=0&sortBy=STATUS&sortDir=ASC) and found that for version 3.6 the cachedResumeToken is initially stored as null. It might be worth including a null check in the if condition for this logic to avoid these scenarios.

Moving this thread here. I agree with adding a null check. I don't expect customers to be using v3.6 in production, but it's still worth avoiding the null pointer exception. Out of curiosity, did you find any valuable details from the logs you've added?

arahmanan (commented):
Final comment before approving!

src/integrationTest/java/com/mongodb/kafka/connect/source/MongoSourceTaskIntegrationTest.java
Calvinnix: Yeah the [logs](https://parsley.mongodb.com/test/mongo_kafka_integration_tests__version~3.6_os~ubuntu_topology~replicaset_javaVersion~JDK17_integration_test_task_patch_d477f00c5506ac461ebc86cbeaaff291598d921c_67116e328067f60007286765_24_10_17_20_06_24/0/9f36feb7f14ab852d4ae3b9e0858a917?bookmarks=0,620&shareLine=0) showed that `cachedResumeToken` was being initialized as `null`. (search for `CALVIN:` if you wanted to review the logs)

I agree, I'll add a null check as well as an additional log so we can see if a null resume token is why we skipped that condition.

Calvinnix: Ok I just added the null check, I'll let the tests run but I do expect the test to still fail for version 3.6 because it doesn't have a resume token to set and resume from.

Calvinnix: Yep they fail in 3.6 for that reason still. https://spruce.mongodb.com/task/mongo_kafka_integration_tests__version~3.6_os~ubuntu_topology~replicaset_javaVersion~JDK17_integration_test_task_patch_d477f00c5506ac461ebc86cbeaaff291598d921c_671653eae106f80007f4ba73_24_10_21_13_15_23?execution=0&sortBy=STATUS&sortDir=ASC

arahmanan (approved):
LGTM!

PR #171, opened 2024-12-06
KAFKA-431: Use Time.SYSTEM for initializing system time

Calvinnix:
Kafka removed support for the `new SystemTime()` instantiation as part of [version 3.9](https://downloads.apache.org/kafka/3.9.0/RELEASE_NOTES.html)

Specifically as part of this [PR](https://github.com/apache/kafka/pull/16266/files).

This change should be safe as far as backwards compatibility goes because that Time.SYSTEM singleton interface [has existed for a long time](https://github.com/dujian0068/kafka/commit/128d0ff91d84a3a1f5a5237133f9ec01caf18d66).

### DIFF
--- a/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
+++ b/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java
@@ -58,7 +58,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 import org.apache.kafka.clients.producer.RecordMetadata;
-import org.apache.kafka.common.utils.SystemTime;
 import org.apache.kafka.common.utils.Time;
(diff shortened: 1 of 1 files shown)


arahmanan (approved):
LGTM! What do you think about filing a ticket to test against different versions of Kafka?

Calvinnix:
> LGTM! What do you think about filing a ticket to test against different versions of Kafka?

Would that do something like automatically download new versions as well? I think that's a good idea

arahmanan:
> > LGTM! What do you think about filing a ticket to test against different versions of Kafka?
> 
> Would that do something like automatically download new versions as well? I think that's a good idea

The first iteration doesn't necessarily have to test new versions automatically. I'm not against it if it's simple enough though

Calvinnix:
https://jira.mongodb.org/browse/KAFKA-432

PR #175, opened 2025-04-10
KAFKA-438: Create DeleteOneTombstoneBusinessKeyStrategy.java

Calvinnix:


src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneTombstoneBusinessKeyStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+import com.mongodb.client.model.WriteModel;
+
+import com.mongodb.kafka.connect.sink.Configurable;
+import com.mongodb.kafka.connect.sink.MongoSinkTopicConfig;
+import com.mongodb.kafka.connect.sink.converter.SinkDocument;
+import com.mongodb.kafka.connect.sink.processor.id.strategy.IdStrategy;
+import com.mongodb.kafka.connect.sink.processor.id.strategy.PartialKeyStrategy;
+import com.mongodb.kafka.connect.sink.processor.id.strategy.PartialValueStrategy;
+
+public class DeleteOneTombstoneBusinessKeyStrategy implements WriteModelStrategy, Configurable {
```
Calvinnix, line 36: I ended up opting to just create a new write model instead of updating DeleteOneBusinessKeyStrategy, the risk would be if we add the ability to start deleting on key values when tombstone events are sent, documents might start being deleted when they weren't in previous versions.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneTombstoneBusinessKeyStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+        document
+            .getKeyDoc()
+            .orElseThrow(
+                () ->
+                    new DataException(
+                        "Could not build the WriteModel,the key document was missing unexpectedly"));
+
+    if (isPartialId) {
+      BsonDocument businessKey = idStrategy.generateId(document, null).asDocument();
+      businessKey = flattenKeys(businessKey);
```
arahmanan, line 52: [q] any ideas on why we flatten the keys only with a partial id strategy? Would it make sense to do the same with a `FullKeyStrategy`?

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneTombstoneBusinessKeyStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+      return new DeleteOneModel<>(businessKey);
+    }
+
+    return new DeleteOneModel<>(kd.containsKey(ID_FIELD) ? kd : new BsonDocument(ID_FIELD, kd));
```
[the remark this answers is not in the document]
Calvinnix: Yeah I agree, originally I had this logic throwing an error saying something along the lines of "DeleteOneTombstoneBusinessKeyStrategy expects PartialKeyStrategy to be defined"

This logic would let us remove defining that kd variable.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneTombstoneBusinessKeyStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Override
+  public void configure(final MongoSinkTopicConfig configuration) {
+    idStrategy = configuration.getIdStrategy();
+    isPartialId =
+        idStrategy instanceof PartialKeyStrategy || idStrategy instanceof PartialValueStrategy;
```
Calvinnix, line 63: Because this is specifically for tombstone events I'm going to remove PartialValueStrategy

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneTombstoneBusinessKeyStrategy.java
Calvinnix: I did this to maintain consistency with the other `*BusinessKeyStrategy`s, I guess technically they could always just use PartialKeyStrategy and just define all of the fields.

src/main/java/com/mongodb/kafka/connect/sink/writemodel/strategy/DeleteOneTombstoneBusinessKeyStrategy.java
```diff
@@ -0,0 +1,63 @@
...
+  @Override
+  public WriteModel<BsonDocument> createWriteModel(final SinkDocument document) {
+    document
+        .getKeyDoc()
+        .orElseThrow(
+            () ->
+                new DataException(
+                    "Could not build the WriteModel,the key document was missing unexpectedly"));
+
+    if (isPartialId) {
```
arahmanan, line 48: [opt] thoughts on returning early for the error case instead? Also, is `isPartialId` still needed? Could we directly check `if (idStrategy instanceof PartialKeyStrategy)`?

arahmanan (approved):
LGTM! just an optional comment.

PR #207, opened 2026-03-24
KAFKA-477: Add github workflow for releasing new versions

Calvinnix:


.github/workflows/release.yml
```diff
@@ -0,0 +1,162 @@
...
+  release:
+    runs-on: ubuntu-latest
+    environment: release
+    steps:
+      - name: Generate GitHub App token
+        id: app_token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
```
Calvinnix, line 29: I will need to create this

.github/workflows/release.yml
```diff
@@ -0,0 +1,162 @@
...
+        run: |
+          sed -i "s/^version = \".*\"/version = \"${{ steps.versions.outputs.release }}\"/" build.gradle.kts
+
+      - name: Commit and push release version
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add build.gradle.kts
+          git commit -m "Version: bump ${{ steps.versions.outputs.release }}"
+          git push origin master
```
Calvinnix, line 99: I will add an exception in the branch protection that let's this workflow push to master.

.github/workflows/release.yml
```diff
@@ -0,0 +1,162 @@
...
+        run: |
+          VERSION=$(grep -oP 'version\s*=\s*"\K[^"]*' build.gradle.kts)
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "Current version: $VERSION"
+
+      - name: Validate version is SNAPSHOT
+        run: |
+          CURRENT="${{ steps.current.outputs.version }}"
+          if [[ "$CURRENT" != *-SNAPSHOT ]]; then
+            echo "::error::Expected a SNAPSHOT version, got: $CURRENT"
```
andrewzlchen, line 52: There's some weird spacing here - or maybe it's expecting to be displayed somewhere that handles ::error::

there are two more places that do this

.github/workflows/release.yml
```diff
@@ -0,0 +1,162 @@
+name: Release
+
+on:
+  workflow_dispatch:
```
andrewzlchen, line 4: so this means that we bump the version via the GHA workflow ui right?

andrewzlchen (approved):
looks great! is this triggered upon making a e

.github/workflows/release.yml
Calvinnix: Yep exactly!

.github/workflows/release.yml
Calvinnix: I'll update the formatting

Calvinnix:
de-prioritizing for now, this would be nice to have in the future though.

